PR #59, opened 2012-12-05
Authentication and CoreAdmin inprovements pysolr.py

w32-blaster:
Added set_credentials method to Solr and SolrCoreAdmin.
Updated Solr._send_request() to use username/password provided
Updated SolrCoreAdmin._get_url() to use username/password provided
Updated SolrCoreAdmin.create(), because the sent action was incorrect (STATUS instead of CREATE)
Updated SolrCoreAdmin.create(), added the optional dataDir parameter

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -305,6 +305,10 @@ class Solr(object):
     def _send_request(self, method, path, body=None, headers=None):
         if TIMEOUTS_AVAILABLE:
             http = Http(timeout=self.timeout)
+            
+            if self.http_user is not None:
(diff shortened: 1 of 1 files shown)


w32-blaster:
Closing after 4 years :|

acdha:
Hey, sorry about that, if you wanted to update this against the current code it'd require some changes since we switched to `requests` – and the auth & SolrCodeAdmin bits should probably be separate pull requests.

w32-blaster:
Well, I think it was useful in 2012 but not anymore. Probably there are better ways to do it now (and with the current code in the master branch)

PR #65, opened 2013-01-31
[FIXED] :: support for http basic authentication

adamcupial-codepill:
Passed auth param to requests, thus allowing http basic authentication

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -7,7 +7,6 @@ import logging
 import re
 import requests
 import time
-import types
 
 try:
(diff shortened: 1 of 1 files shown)


evalapply:
Hello, There are at least two old unresolved pull requests with authentication. There is no discussion. Are they unresolved only because of conflicts? Would you welcome another pull request with this feature?

acdha:
@evalapply I would highly welcome a pull request against master with some tests and a doc update.

PR #67, opened 2013-02-06
Add support for fetching term vectors via the tvrh request handler.

adityar7:


mbeacom:
What's the status of merging in @adityar7's term vectors functionality?

PR #73, opened 2013-02-26
SolrCoreAdmin nonfunctional

Jaidan:
SolrCoreAdmin as far as I could tell is completely non functional as it stands currently.  None of the methods work as written and all return the same result as you would get for /solr/admin/cores with no parameters.

This patch makes all SolrCoreAdmin methods functional, adds the missing data_dir parameter to create, and implements tests that actually assert that the SolrCoreAdmin is doing anything at all.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -203,24 +203,8 @@ class Results(object):
     def __iter__(self):
         return iter(self.docs)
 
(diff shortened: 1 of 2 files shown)


tisto:
+1

PR #78, opened 2013-03-26
Added support for realtime get

anti-social:
See: http://yonik.com/solr/realtime-get/

With this feature it is possible to use Solr as NoSQL database.

This pull-request https://github.com/toastdriven/pysolr/pull/74 doesn't support ids argument. Also Solr supports only id and ids parameters.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -608,6 +608,19 @@ class Solr(object):
         return Results(response.get('docs', ()), numFound, **result_kwargs)
 
     def get(self, id=None, ids=None, **kwargs):
+        """
+        Gets latest version of documents, even uncommited ones.
+
+        Requires *either* ``id`` or ``ids``. ``id`` for single document.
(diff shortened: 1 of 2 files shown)


joekiller:
This is a pretty old PR but would be nice to have @acdha.

acdha:
@joekiller Let's dust it off then – have you had time to test this pull-request?

joekiller:
I monkey patched the methods in and they appear to work fine.  Will try to
give the PR a spin.

joekiller:
I checked out the branch and ran the tests. Four failed but they are not related to the PR.  Also I was running solr 4.10 which may have made a difference.  The failing tests were:

  * `test__mlt`
  * `test__scrape_response`
  * `test_extract`
  * `test_more_like_this`

I've used the methods and they work as expected.  I'd give this a :+1:

PR #92, opened 2013-08-05
fixed lxml errors when reading Tomcat error messages. 

atkinson:
When parsing error messages pysolr assumes that Tomcat will send a certain flavour of invalid response.

Sometime in Tomcat 6 (or maybe Solr4) the assertion that this code was based on became untrue, and so the error handling code in pysolr began creating it's own error. This may only be true when using lxml (it's required in my project so I haven't tested without).

This fix prevents pysolr from obscuring the tomcat error message with it's own, if it fails to find the tag it's looking for.

This is what I was getting before making this fix:

    File "/home/webapps/.virtualenvs/myapp/local/lib/python2.7/site-packages/pysolr.py", line 404, in _scrape_response
    p_nodes = body_node.cssselect('p')
    AttributeError: 'NoneType' object has no attribute 'cssselect'

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -426,25 +426,28 @@ class Solr(object):
         dom_tree = None
 
         if server_type == 'tomcat':
-            # Tomcat doesn't produce a valid XML response
-            soup = lxml.html.fromstring(response)
-            body_node = soup.find('body')
-            p_nodes = body_node.cssselect('p')
+            try:
+                # Tomcat doesn't produce a valid XML response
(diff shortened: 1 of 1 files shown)


pembo13:
What's the status of this, I'm still getting errors when trying to build an index.

acdha:
@pembo13 If you can, please test the branch I referenced above – and send us some of the Tomcat error messages you're receiving so we can add them to the test suite.

domenkozar:
This can be closed now, code no longer exists.

PR #94, opened 2013-09-13
Make _to_python recursive for lists and tuples

tuky:
`_to_python`'s behavior on lists / tuples to only convert the first value seems very arbitrary. It even fails with an `IndexError` on empty lists. This PR also fixes multivalued dynamic fileds in e.g. Solr with haystack: they have no convert function and therefore call this `_to_python` method.


tuky:
It would still be nice to see this upstream. Does anything speak against this change?

acdha:
It does look like a reasonable change, although it'd need a corresponding change for consumers like e.g. Haystack. That might be worth cutting 4.x.

cclauss:
Please rebase.

PR #121, opened 2014-07-03
Support for termVectors

dmugtasimov:
I need to access termVectors of TermVectorComponent:
https://wiki.apache.org/solr/TermVectorComponent


AlJohri:
:+1:

acdha:
I'd like to move away from so many custom accessors which require pysolr to closely track Solr's API changes (see #195). How valuable is this versus simply using `.raw_response['termVectors']`?

PR #126, opened 2014-08-23
raise here if a body node wasn't found

yeago:


domenkozar:
This can be closed now.

yeago:
did it or anything like it make it?

domenkozar:
yes, now regex is used to extract messages. Solr 4 is also supported.

PR #132, opened 2014-10-16
Update to specify lxml as required for Tomcat

frankamp:
Lxml is not so much optional as required under certain circumstances. I added the tag 'required for Tomcat error support' to cssselect and lxml because lxml is imported with no try/except when the solr server is running on Tomcat.


domenkozar:
This can be closed now.

PR #133, opened 2014-10-16
Remove lxml requirement from error parsing

frankamp:
LXML is overkill and not really required to make a decent error string. This is my take at a reasonable alternative to the htmlsoup parsing, removing a couple dependencies. While the htmlsoup parsing could be considered flexible it is still much more dependent on present form html structure than need be. I think the H1 will be long term more reliable error string than the more complicated pathing required of the current algorithm. Either way we are betting on the future output of an external project (solr + tomcat), and writing pretty fragile code.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -424,7 +424,6 @@ class Solr(object):
             server_type = 'jetty'
 
         if server_string and 'coyote' in server_string.lower():
-            import lxml.html
             server_type = 'tomcat'
 
         reason = None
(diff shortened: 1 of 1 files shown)


acdha:
@frankamp Did you have a chance to review that code? I was updating the docs for pysolr & haystack and realized that it'd be nice to remove the lxml dependency as well.

frankamp:
Sorry I didn't have a chance to get back to this. It looks like this can close out now that the work is over on #149

Issue #51, opened 2012-06-06
pysolr should support custom handlers


llnik:
Solr provides the creation of various handlers for different purposes in solrconfig.xml - so, one could have a "/solr/select handler for default searching and a /solr/articlesearch for searching with some different filters and different settings.

In pysolr, however, only the default handlers (/solr/select, /solr/update, etc.) can be used.

mbeacom:
@toastdriven @acdha Are you still actively merging pull requests?  Can we please get some of these patches merged in or explanation as to why they were rejected?  Thank you!

acdha:
@mbeacom If you need either of these, the first place to start would be updating the code against the current branch to confirm that there are no test regressions. I work on pysolr in my spare time so anything which makes review easier is going to help.

mbeacom:
@acdha I am not trying to add more work for you.  I'll go ahead and get it up to speed.  Thank you!

mbeacom:
@acdha https://github.com/toastdriven/pysolr/pull/161

nuarhu:
I've rebased (again) onto the current master. Is there a problem or are there considerations that prevent you from merging this? We are using this in production without issues, so far.

PR #134, opened 2014-10-21
Request handler

nuarhu:
This is a pull request that contains and replaces #112 , addressing #51 and #52 :

Additionally to specifying request handlers at the method level, it allows to specify a `search_handler` as instance variable. Also, a new instance variable `use_qt_param` is available that allows to change how request handlers are specified in the SOLR URL. The default SOLR behavior is to expect the handler as URL path variable (current #112 implementation). The pre 3.6 default behavior (and still available via configuration) specifies the request handler via `qt` parameter in the URL.

This pull request will allow to specify a custom request handler at construction time and thus allow to configure it as data source for Haystack like this:

    HAYSTACK_CONNECTIONS = {
        'default': {
            'ENGINE': 'haystack.backends.solr_backend.SolrEngine',
            'URL': 'http://127.0.0.1:8983/solr/core0'
        },
        'autocomplete': {
            'ENGINE': 'haystack.backends.solr_backend.SolrEngine',
            'URL': 'http://127.0.0.1:8983/solr/core0',
            'KWARGS': {
                'search_handler': '/autocomplete',
                #'use_qt_param': False,  # False by default
            }
        },
    }

Update README to include documentation of custom handlers.

I had to change the import statement in `get-solr-download-url.py` to make the tests run. (Using 3.4 but actually my fix should be for python3 in general.)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -115,21 +115,26 @@ Simply point the URL to the index core:
     solr = pysolr.Solr('http://localhost:8983/solr/core_0/', timeout=10)
 
 
-Custom Request Handler
-~~~~~~~~~~~~~~~~~~~~~~
+Custom Request Handlers
+~~~~~~~~~~~~~~~~~~~~~~~
 
 .. code-block:: python
 
     # Setup a Solr instance. The trailing slash is optional.
-    solr = pysolr.Solr('http://localhost:8983/solr/core_0/', query_handler='/autocomplete', use_qt_param=False)
+    solr = pysolr.Solr('http://localhost:8983/solr/core_0/', search_handler='/autocomplete', use_qt_param=False)
 
 
 If ``use_qt_param`` is ``True`` it is essential that the name of the handler is exactly what is configured
-in solrconfig.xml, including the leading slash if any (though with the qt parameter a leading slash is not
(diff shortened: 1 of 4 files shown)


nuarhu:
I've rebased this onto toastdriven/master.

nuarhu:
Rebased onto current master.

nuarhu:
The following test run fails:

    Python 3.3.; SOLRCLOUD=False

During all of the tests in this run, no connection can be established to SOLR. This works in all the other test runs. It might be a one time thing?

    Failed to establish a new connection: [Errno 111] Connection refused

I assume this means that might tests are fine, all over, seeing that all of the other runs finish just fine.

acdha:
@nuarhu After restarting that build, it passed – we definitely see sporadic failures caused by how long it can take Solr to startup. I actually just made https://github.com/django-haystack/pysolr/commit/05f0674f30b2641067a3effa969e8dd9c0774df4 to reduce that, although it wouldn't have helped with this particular problem.

nuarhu:
@acdha Thanks for checking! I hope everything is fine with my rebase and you are fine merging it.

nuarhu:
rebased - again! Please consider merging this?

PR #138, opened 2014-11-27
Feature/solrcloud

upayavira:
An extension to pysolr to make it Zookeeper/SolrCloud aware. This is cloned from the code in the SolrJ client.
The tests are limited to proving that this does not break existing functionality, although I have tested (manually) that it does correctly failover between nodes when a node in a cluster fails.
# Commit Checklist
- [ ] Test coverage for ZooKeeper / SolrCloud error states
- [x] Add a Travis test matrix which runs _without_ Kazoo installed to confirm that nothing breaks for traditional usage (the SolrCloud tests are supposed to be skipped)
- [ ] [Support SolrCloud 5](https://github.com/toastdriven/pysolr/pull/138/#issuecomment-183096888) and have a Travis test matrix entry for both major versions
- [x] Add test that confirms that Pysolr fails-over correctly when one of the Solr nodes disappears (can be simulated with kill -STOP and kill -CONT)


danizen:
+1 - I need this in my environment.

upayavira:
Any suggestions as to what is holding this patch back from a merge? Thx!

(I notice there's conflicts which I will resolve)

mmroden:
I would really like this feature to be merged in; what needs to be done to make that happen, apart from conflict resolution?

acdha:
Beyond conflict cleanup & a quick style review, the main thing I'd want to see would be updating the test runner so it continues to test regular Solr along with running an instance with SolrCloud enabled.

The install packaging also might want a little review – this was rolled into the Tomcat target but seems like it should be a separate one, particularly since the Solr project has itself deprecated deployment on Tomcat.

acdha:
That said, if you have time to work on this, I can review & ship a release.

upayavira:
I can look into that, @acdha

acdha:
@upayavira awesome, thanks!

upayavira:
@acdha I've made some reasonably substantial changes to this PR. Basically, instead of starting Solr and getting on with it, we now need the tests to start/stop the correct type of Solr, so I improved your start-solr-test-server.sh script to prepare and start both non-cloud and cloud instances, and then extended the tests to make them able to start the right Solr in their setupClass method.

By extending the SolrTestCase with a SolrCloudTestCase, I was able to have most of your existing tests run a second time against a solrcloud instance.

The tests seem to consistently pass, but for some reason seem to hang at the end, for reasons I have yet to identify.

I am considering rewriting start-solr-test-server.sh in Python, but more importantly I want/need a test that hits Solr repeatedly whilst killing one of the two nodes that contains the collection being queried. Note, it starts three Solr nodes, one for Zookeeper and two to host collections. This means we can kill either of the collection nodes without disrupting Zookeeper.

Please let me know what you make of this (rather substantial) PR. Comments/suggestions welcome, before I start writing the above, more complex, test.

mmroden:
For what it's worth, the tests work on my machine (macos 10.10.5) once I change to the feature/solrcloud branch and I install kazoo.  I have these messages, but I think they're in kazoo, not pysolr:

    Waiting for simple-solr ----No handlers could be found for logger "kazoo.client"
    No handlers could be found for logger "kazoo.client"

acdha:
@upayavira I've made some changes in https://github.com/upayavira/pysolr/pull/1. I definitely agree that we're probably close to the point where we should just port the server launcher to Python since we have a bunch of daemons to manage now and will want things like timeouts for hard-kills on the various processes.

acdha:
The main problem I'm seeing is that KazooClient may leak connections which will block shutdown endlessly trying to reconnection. I don't have time to finish debugging this but it did lead to a lot of cleanup.

acdha:
https://github.com/acdha/pysolr/tree/138-solrcloud now has stable tests. The biggest change was in https://github.com/acdha/pysolr/commit/67dcc640d1b668495cfd39eae53495363cb9498a which changed the way we start and stop Solr/SolrCloud to avoid spending most of our test runtime starting and stopping them and prevent tests from hanging where the tests all passed but the KazooClient atexit hacks prevent the system from exiting until they can restore a connection to the ZooKeeper instance which was already shut down by that point.

upayavira:
The starting and stopping of Solr for each test case was intentional - as another test is needed - one that confirms that calls to Solr still work once a node has been shut down.

However... I've since worked out that, if we assume we are on Unix, we could achieve that by using kill -17 $PID to stop a Solr instance (equivalent to CTRL+Z at a terminal) and kill -19 to start it again. That would actually be compatible with having both nodes started at the beginning of the test run, and stopped at the end, so your simplification should actually be fine.

I have some other improvements that I made before I saw your work on this - I'm currently working on merging them, which is somewhat tough because we both touched very much the same stuff.

acdha:
Yes – sorry about that. I strongly prefer the approach of using `kill -STOP` / `-CONT` on an instance or starting a third one just for those tests because starting and stopping Solr is bad enough even without restarting ZooKeeper a few times and I'd really like to avoid performance discouraging people from running the tests.

upayavira:
Absolutely - I now see the kill -STOP thing is way better, but I only worked that out this morning - before that I was assuming I would need to start/stop instances. On other words - nothing to apologise for - I'm appreciating your work.

acdha:
Right now, I'd like to get the excluded failures resolved and review the coverage impact for this.

I've been seeing a bunch of other things which I'd like to fix which aren't specific to this branch so I'd like to plan to merge this fairly soonish so we can get ready for a 4.0 release with some of the hairier old bits of code removed.

upayavira:
One thing I should note - we are testing this against 4.10.4. I'm not sure when it happened, but I suspect after that - Solr moved away from a single clusterstate.json to having a file per collection. Thus, this code isn't compatible with Solr instances in the 5.x range - we will need to have that handled before a major release as it *will* be expected. I don't envisage that to be hard, but it will need doing.

acdha:
https://github.com/acdha/pysolr/commit/d26016b37bb801864017aa11d71066052d491cab gets some of the remaining tests flagged as expected failures but might want to be a little cleaner in how it handles the core names.

acdha:
Agreed on the need to handle both versions. It would probably be a good idea to think about how to handle this – I'm initially disinclined to have `SolrCloud4` and `SolrCloud5` classes if we can just use something like a `solr_version` constructor option or something like that but maybe there's something to be said for putting it into the class name to make it explicit.

acdha:
As of right now, we're at ~83% test coverage with most of the uncovered code being in `getAliasHosts`

acdha:
I added a simple checklist points to the description on this ticket. Feel free to add more as they come up.

If we reach the point where this seems stable before getting to the more complicated tests, I could merge the working branch to master & we can start new pull requests from that point.

upayavira:
From my understanding of the changes in Solr 5, we should be able to make it handle both - it watches both locations and adapts depending on what it finds in Zookeeper.

I agree with your plan to stabilise what we have and merge it before looking at additional tests/features.

acdha:
@upayavira I'm taking a bit of a break from this today if you want to make some changes without worrying about us stepping on each other. I'd like to have something ready to merge by next week so we can polish it a bit. It'd be really nice to see if we could hit 100% test coverage for a release.

upayavira:
I presume by 100% test coverage you mean that all tests pass? That should now be the case. I also raise a few more explicit SolrErrors in there too. Let me know what you think.

acdha:
@upayavira I was thinking that we have 100% coverage of the new code by the test suite. It looked like most of the remaining gaps were in error-handling, which is definitely easy to miss in future changes.

acdha:
master now has a few minor updates, the most notable being https://github.com/acdha/pysolr/commit/d23de5825be37a039e2946ecda056a3e723e4954 to ensure that the byte-strings containing JSON are decoded first before passing them to `json.loads`

acdha:
https://github.com/acdha/pysolr/commit/cbc07af56c6a32ead3e90efbc0cdb5bf13cd2e05 adds an `env:` option to Travis CI so the tests are run with and without Kazoo installed, with the ~32 SolrCloud tests skipped in the latter case.

upayavira:
I have some initial/prototype code that should work with both 4.x and 5.x for SolrCloud, and also implements some of the Collections API for creating/deleting collections/etc. I'll post it as a fresh PR once I've played with it a bit more. Really though, as with the above code, it will need some decent tests around it to prove it behaves properly in failover scenarios.

mmroden:
Hi all,

Where do we stand on this?  Is there a sense that pysolr 4.0 will be coming out soon-ish (ie, next two weeks), or should I go ahead and use this branch as a one-off in a private pypi?

Thanks!

vvolkman:
+1 I really need the custom search_handler in a released version.  Tired of pulling from the tip instead of the pip.  thanks in advance!

acdha:
I don't think we're ready for 4.0 yet unless someone has time to help add tests – in particular, it'd be good to start testing with Solr 5 as well as 4.x.

upayavira:
I have a test for failover which seems to work (it passes). I need to add one in which both nodes are down and confirm that it fails. 

This code WILL NOT work with Solr 5.x. I've got as far as working out exactly what we need to do (and have some hacked, unpublishable code), but just need to find the time to make it happen.

danizen:
Upayavira, can you expand on where it won't work?   I have at work pysolr
working with Solr 5 in a non-cloud mode, and I'm anticipating a switch to
cloud.

I have a test for failover which seems to work (it passes). I need to add
one in which both nodes are down and confirm that it fails.

This code WILL NOT work with Solr 5.x. I've got as far as working out
exactly what we need to do (and have some hacked, unpublishable code), but
just need to find the time to make it happen.

—
Reply to this email directly or view it on GitHub
<https://github.com/django-haystack/pysolr/pull/138#issuecomment-193972263>.

upayavira:
When Solr 4.0 released, SolrCloud stored its cluster information in clusterstate.json - every collection in a single file in Zookeeper. This worked, but people started building clusters with 1000s of collections, leading to an unmanageable file. Therefore, somewhere around the 5.0 mark it switched to having a per-collection state.json file. This allows a zookeeper client to follow just the collections it is interested in.

I don't see any reason why the client cannot be coded to handle both setups. We would need a way to test against both, though, which might require two different Solr installs to run our tests against.

upayavira:
See https://github.com/django-haystack/pysolr/pull/187 for further work here. Should we close this PR now?

upayavira:
> Test coverage for ZooKeeper / SolrCloud error states

If you can clarify what you are looking at here, I can look into that.

> Support SolrCloud 5 and have a Travis test matrix entry for both major versions
Solr 5.x support is in PR #187 . 

I plan to pull the start-solr-test-server.sh for 4.10, rename it start-solr-test-server-4.10.1.sh, and add a way that run-tests.py can use one or the other. That'll mean we can have a Travis run for each version, both using the latest (5.x compatible) code.

> Add test that confirms that Pysolr fails-over correctly when one of the Solr nodes disappears (can be simulated with kill -STOP and kill -CONT)

This test is, again, in PR #187 . I accept that I should have separated PR 187 into three PRs, but they kinda all happened at the same time. If you want them separating, let me know and I'll see how possible that is.

acdha:
@upayavira The main thing I was thinking is confirming that we do something reasonable when ZooKeeper itself has either failed outright or is unresponsive. I think that'd just be a question of making sure our timeouts are used and that something useful is logged.

Since you have a test in #187 I checked the failover tests off the list here.

upayavira:
I've been working all morning on the failover tests. Getting them right isn't easy. It seems that a kill -17 does not cause a change the live_nodes entry in ZooKeeper, so we cannot use a kill -17.

Test scenarios:
 * Does it handle when one replica is down for a shard of a collection?
 * Does it fail gracefully when there are no nodes available for a shard/collection?
 * Does it fail gracefully if it cannot connect to ZK, and recover once ZK comes back?

These are non-trivial tests. The code under test is WAY simpler. I'm mulling now on how to implement them, but I'm open to suggestions,

acdha:
One thing I've been wondering is whether it makes sense to merge the work-in-progress here and postpone some of the more involved tests or simply defer them entirely to Kazoo or the upstream ZK/Solr projects and instead just use something like `mock.patch` to confirm that our logging code, etc. works when the important KazooClient methods raise exceptions.

upayavira:
See code in #187 , just committed. I believe it does the job for the above three test scenarios. It turns out it was important to get the tests right, as they did show up one particular bug (not readding a watch when a watch is triggered).

All that is left is support for both 4.x and 5.x. And, by comparison, that is seriously trivial!

upayavira:
I've now got code that works against both 4.x and 5.x. You can set an environment variable to tell it which startup script to use (which version of start-solr-test-server.sh), i.e. the 4.x or the 5.x one. Thus, it should be possible to keep a version of each running in Travis.

This also involved redoing the retry logic somewhat. Solr 4.10 wasn't updating its cluster state as fast as it should, so I made it, by default, retry every 0.02s, and retry 20 times. I saw maybe 4 or 5 retries needed in my tests. These can be overridden on the SolrCloud constructor. I also needed to call random.seed() as random.choice was giving me the same value every time.

I will commit this work against #187 as soon as I get a next moment.

acdha:
@upayavira that's excellent news!

upayavira:
Moments have been far apart, but #187 now has my latest changes. You can run the tests against 5.5 with:
    ./run-tests.py
or against 4.10.1 with:
    PYSOLR_STARTER=./start-solr-test-server-4x.sh ./run-tests.py
Add both of these to Travis, and you should have good test coverage.

PR #139, opened 2014-11-27
Optimize is no longer recommended

upayavira:
Since Solr 3.6, Solr has used the TieredMergePolicy which makes,
in most scenarios, optimization a harmful rather than beneficial
step.

We should leave the feature in the client, but not publicize it, IMO.


toastdriven:
Thanks!

PR #142, opened 2014-12-02
added cursormark deep pagination support

yspanchal:
Added cursormark solr deep pagination support


swistakm:
IMO it's quite useful feature. Why it has not been merged yet?

PR #148, opened 2015-02-12
Preserve maxScore in the results object.

alexef:
I'm using maxScore for debugging purposes, it seems that it is lost currently in the Results wrapper.


alexef:
No activity.

acdha:
@alexef We need a documentation update – v3.5 added a `raw_response` property which you can use to get fields like this without needing to maintain an ever-growing list of properties:

https://github.com/django-haystack/pysolr/commit/329896cef73a3fd929aa4a415d02c6c75fa0ff43

Issue #159, opened 2015-06-06
Not 200 solr response cause crash on python3


touilleMan:
Hi there,

I've encountered an encoding error (see: https://github.com/toastdriven/pysolr/blob/master/pysolr.py#L445)

```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/app/.heroku/python/lib/python3.4/site-packages/pysolr.py", line 813, in add
    return self._update(m, commit=commit, softCommit=softCommit, waitFlush=waitFlush, waitSearcher=waitSearcher)
  File "/app/.heroku/python/lib/python3.4/site-packages/pysolr.py", line 399, in _update
    return self._send_request('post', path, message, {'Content-type': 'text/xml; charset=utf-8'})
  File "/app/.heroku/python/lib/python3.4/site-packages/pysolr.py", line 328, in _send_request
    error_message = self._extract_error(resp)
  File "/app/.heroku/python/lib/python3.4/site-packages/pysolr.py", line 417, in _extract_error
    reason, full_html = self._scrape_response(resp.headers, resp.content)
  File "/app/.heroku/python/lib/python3.4/site-packages/pysolr.py", line 445, in _scrape_response
    if response.startswith('<?xml'):
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
```

Basically, the issue is `response` (i.e. response's content) is made of bytes where `'<?xml'` is an unicode string on python3.
We could solve this pretty easily by prefixing the string with a `b` to indicate it is supposed to be a bytes string under python3 (and a regular str under python2)

Do you want me to provide this pretty simple fix ?

DrChai:
It happens on Python 3.4.3 as all I known (i just tested on python 3.4.0 and 3.4.3, it works for v3.4.0).
Reason is the response variable is bytes not a string. you have to change the line 445 to:
response.startswith(b'<?xml'): // makes '<?xml' in bytes
and then call
sudo python setup.py install

touilleMan:
@DrChai I think you should provide a (pretty simple) PR for this issue (but if you don't have time for this I can do it)
@toastdriven Do you agree about a PR ?

DrChai:
@touilleMan you could just do PR for this. 
And i was wrong about its working on 3.4.0. I just realized it just didnt have a chance to call that function.
More details u can take a look at this answer:
http://stackoverflow.com/questions/6269765/what-does-the-b-character-do-in-front-of-a-string-literal/6269785#6269785

[...]

PR #162, opened 2015-07-18
Fix unittests & solr response error handling on python3

touilleMan:
Hi there !

Following issue #159, I went working a bit on this project

 - First I had to fix the solr installer to allow travis to run the tests
 - Then I fix issue #159 and add an unittest test about it
 - Finally, I had to mark as expected faillure `test_extract` given it was weirdly failing under travis

Here is the stack trace of the error (or see https://travis-ci.org/touilleMan/pysolr/jobs/71444801, the error is present both on python 2.x and 3.x)

```
======================================================================
ERROR: test_extract (tests.client.SolrTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/client.py", line 513, in test_extract
    extracted = self.solr.extract(fake_f)
  File "pysolr.py", line 932, in extract
    files={'file': (file_obj.name, file_obj)})
  File "pysolr.py", line 331, in _send_request
    raise SolrError(error_message)
SolrError: [Reason: lazy loading error]
```

My guess is travis is missing a library, but I couldn't figure really what...

### DIFF
--- a/tests/client.py
+++ b/tests/client.py
@@ -194,6 +194,11 @@ class SolrTestCase(unittest.TestCase):
         self.assertRaises(SolrError, self.solr._send_request, 'get', 'select/?q=doc&wt=json')
         self.solr.url = old_url
 
+        # Test bad core as well
+        self.solr.url = 'http://localhost:8983/solr/bad_core'
+        self.assertRaises(SolrError, self.solr._send_request, 'get', 'select/?q=doc&wt=json')
+        self.solr.url = old_url
+
     def test__select(self):
         # Short params.
         resp_body = self.solr._select({'q': 'doc'})
--- a/tests/client.py
+++ b/tests/client.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import datetime
 import sys
+from os import environ
 
 from pysolr import (Solr, Results, SolrError, unescape_html, safe_urlencode,
                     force_unicode, force_bytes, sanitize, json, ET, IS_PY3,
@@ -498,6 +499,7 @@ class SolrTestCase(unittest.TestCase):
         self.solr.optimize()
         self.assertEqual(len(self.solr.search('doc')), 4)
 
+    @unittest.skipIf(environ.get('TRAVIS'), reason=u"Temporary fix for this strange error with travis")
     def test_extract(self):
         fake_f = StringIO("""
             <html>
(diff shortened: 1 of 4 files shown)


elishowk:
It would be great merging and releasing a new version, downstream some django-haystack PR are blocked because of this issue https://travis-ci.org/django-haystack/django-haystack/jobs/76260939

noisy:
looks ok :+1:

vinnyrose:
:+1:

tests/client.py
```diff
@@ -194,6 +195,11 @@ def test__send_request(self):
         self.assertRaises(SolrError, self.solr._send_request, 'get', 'select/?q=doc&wt=json')
         self.solr.url = old_url
 
+        # Test bad core as well
+        self.solr.url = 'http://localhost:8983/solr/bad_core'
+        self.assertRaises(SolrError, self.solr._send_request, 'get', 'select/?q=doc&wt=json')
+        self.solr.url = old_url
```
acdha: This should be a try/finally block to avoid the possibility of an unanticipated error leaving the test `self.solr.url` value in place

touilleMan:
I've corrected my branch with the try/finally block :+1:

acdha:
The Solr errors turned out be caused by a change in file paths between Solr releases – the path in `solrconfig.xml` used to load the content-extraction library (solr-cell) from `contrib/extraction/lib` was one level off. I changed it to define the `solr.install.dir` property in the hopes that this will avoid similar problems in the future:

https://github.com/toastdriven/pysolr/commit/16c993459de0139f007bcd9ed35ba15b9eac16cf#diff-9588af6c9d46310c2f216d9e14a0b959R55

acdha:
After that change, the only commits left on this PR were squashed into https://github.com/acdha/pysolr/commit/cf0fa493ea73217a427475a5c6a88fcf65220168 which ran successfully:

https://travis-ci.org/acdha/pysolr/builds/82057531

touilleMan:
cool :+1: 

The only remaining trouble is the travis status button on the README which points on a non existing target (https://travis-ci.org/toastdriven/pysolr)

acdha:
Yes – unfortunately, only @toastdriven can check the box to enable that CI build. I'd prefer to transfer this repo to the django-haystack organization so it doesn't depend a single person.

acdha:
cf0fa493ea73217a427475a5c6a88fcf65220168 was merged awhile ago but I just added some credits to the AUTHORS file in bf6d8b648738c575c9067b765b6f233c4921e7dd.

If you have any testing or suggestions for the README, I'm getting ready for the v3.4.0 release.

PR #165, opened 2015-09-11
Refactor start-test-solr.sh to support Solr 4 and Solr 5

crantila:
The test suite now uses a single-core example that's present in both Solr 4 and 5. The script detects the appropriate Solr version at runtime and prepares a Solr home directory accordingly. Some of the tests, which assumed the core is called "core0," have been modified to use a core called "collection1."

To be clear: the test suite runs with Solr 5, but it doesn't yet pass.

Also, the previous home directory used a two-core Solr configuration. I couldn't see anywhere that both cores were being used, and Solr 5 doesn't seem to have a corresponding "multicore" example, so this just uses the default single-core configuration now. If there's a real need for the multicore test configuration, or if you'd like me to make other changes, I'd be happy to revise my pull request!

I will be making the necessary changes for the test suite to pass with Solr 5.

### DIFF
--- a/start-test-solr.sh
+++ b/start-test-solr.sh
@@ -4,6 +4,11 @@ set -e
 
 SOLR_VERSION=4.7.2
 
+if [ ! -d "test-solr-server" ]; then
+    mkdir "test-solr-server"
+fi
+cd "test-solr-server"
+
 export SOLR_ARCHIVE="solr-${SOLR_VERSION}.tgz"
 export SOLR_DIR="solr-${SOLR_VERSION}"
 
--- a/start-test-solr.sh
+++ b/start-test-solr.sh
@@ -4,9 +4,7 @@ set -e
 
 SOLR_VERSION=4.7.2
 
-if [ ! -d "test-solr-server" ]; then
-    mkdir "test-solr-server"
-fi
+mkdir -p "test-solr-server"
 cd "test-solr-server"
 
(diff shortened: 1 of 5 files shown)


start-test-solr.sh
```diff
@@ -1,11 +1,13 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
 SOLR_VERSION=4.7.2
 
-export SOLR_ARCHIVE="${SOLR_VERSION}.tgz"
+export SOLR_ARCHIVE="solr-${SOLR_VERSION}.tgz"
+export SOLR_DIR="solr-${SOLR_VERSION}"
```
acdha: What do you think about changing this to make everything happen inside a directory like `test-solr-server`? That way we could have a single pattern for the .gitignore, prefix everything in here, etc.

acdha:
Thanks for starting this process moving – this looks like a good start since we're going to have to maintain the ability to test with Solr 4.x and 5.x for awhile while everyone rolls through the upgrade process

start-test-solr.sh
crantila: Good idea, it would be much cleaner that way. And thanks for responding so quickly!

crantila:
I've made the change to ``mkdir -p`` as suggested above. Is it possible to merge this pull request now?

Issue #166, opened 2015-09-24
Assumptions fail and leaky abstractions


swistakm:
I use pysolr because there is nothing better yet. Sadly every time I need to do something unusual I'm stuck because pysolr makes some painful assumptions about what I wan't to do. Most of all it makes assumptions about what parts of response I'm interested in and does not give me anything more.

Currently we can access only those fields that:
1. are accessed as attributes of `Results()` class
2. `Solr()` class API methods kindly pass to `Results()` class as instance args/kwargs

This approach has following problems:
1. This isn't future-proof. Whenever something new is added to Solr, pysolr users can't access it because assumption was made ("you ain't gonna need that") and can't do nothing about that. Only solution is to add a new "feature" as a PR that lands in limbo (#140, #141, #148, #142, #55; note that some of them are years old!).
2. There is no separation of concerns. `Results()` do nothing more than storing fields. It really could be a dict. There is almost no benefit in using object here. And most of the code in methods like `search()` is creating a result field by field:
   
   ```
   if result.get('stats'):
       result_kwargs['stats'] = result['stats']
   ```
   
   This IMHO should be a responsibility of `Results()` class - take a raw response an make it something more python-like. As a side note i feel that the need for altering this response is doubtful.
3. Cognitive burden - result attributes are different from those used in Solr (like: `qtime` vs. `QTime`). So if you are well familiar with Solr you need to remember two sets of results fields. Also if someone wants to switch to `pysolr` due to other features then migration of existing code can be tedious because of that. 
4. Introducing new solr features requires new tests while it actually should not. Since Solr is quite mature project there is no need to tests things that are documented and outside of the lib scope. But since every not covered response field becomes actual `Results()` class and `Solr()` class code it becomes pysolr's feature and thus needs testing. Take for example the `cursorMark` proposed by #140 & #141.
5. Too much maintenance. Every new field needs new code. Trivial or not this creates need for PRs, reviews, merging, releases. No one likes doing that for trivial code so this is the reason of so long PR backlog piling up on GitHub.

But hey! pysolr is generally good but not "future proof" and thus needs constant care from maintainers. So much that they lost the steam and this is the reason than mentioned PR-limbo exists. This isn't a vicious rant. I have suggestion how to fix this situation in a few steps:
1. quick hotfix so impatient devs (like me) can still can use pysolr despite the fact it does not assumed there are features they would like to use. Solution is simple: make `Results()` object contain also a raw solr response dictionary. Like:
   
   ``` !python
[...]

PR #167, opened 2015-09-25
move response manipulation responsibility to Results class and allow custom results classes to be defined

swistakm:
Reasons for proposed changes are explained in #166

Generally what have changed:
* response manipulation responsibility moved from `.search()`and `.more_like_this()` methods moved to  `Results()` class. This also reduced code duplication between these methods.
* custom results class can be defined for `Solr` instance
* `ResultsTestCase` changed to reflect changes `Results` object creation
* added test for custom results class in `SolrTestCase`

Also as a bonus:
* added missing docstring for `Results()` class that by the way explains how to iterate over documents

I resigned from storing additional raw response dict inside Results because it would at least double this object memory size. If someone needs things that are not covered yet then he can easily create custom results class and how to do that

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -219,19 +219,70 @@ class SolrError(Exception):
 
 
 class Results(object):
-    def __init__(self, docs, hits, highlighting=None, facets=None,
-                 spellcheck=None, stats=None, qtime=None, debug=None,
-                 grouped=None, nextCursorMark=None):
-        self.docs = docs
-        self.hits = hits
-        self.highlighting = highlighting or {}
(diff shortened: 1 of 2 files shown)


skirsdeda:
+1

swistakm:
bump :)

acdha:
I'm getting ready to tag v3.4.0 if you have any suggestions for updates to the README

swistakm:
For me the docstring of Solr class is enough. Also I expect that custom result classes are the feature that only a minority of advanced users will use so I don't know if it is worth to be exposed in README.

PR #168, opened 2015-09-28
Rollback

RomanSek:
Added rollback command.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -897,6 +897,24 @@ class Solr(object):
 
         return self._update(msg, waitFlush=waitFlush, waitSearcher=waitSearcher)
 
+    def rollback(self, waitFlush=None, waitSearcher=None):
+        """
+        Tells Solr to cancel all add/deletes made to the index since the last
+        commit.
+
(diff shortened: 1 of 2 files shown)


pysolr.py
```diff
@@ -897,6 +897,24 @@ def optimize(self, waitFlush=None, waitSearcher=None, maxSegments=None):
 
         return self._update(msg, waitFlush=waitFlush, waitSearcher=waitSearcher)
 
+    def rollback(self, waitFlush=None, waitSearcher=None):
+        """
+        Tells Solr to cancel all add/deletes made to the index since the last
+        commit.
+
```
swistakm: I think that there should be also an clear warning that this is an expert level API and rollbacks may be used only if application takes whole responsibility for updates concurrency, sharing and replication. This is what Solr documentation says about this command.

Developer unexperienced with Solr may think that commits and rollbacks are analogous to how transaction in RMDBS work but that is not true. Also as far as I remember example Solr configurations have often auto commit enabled and this should be always disabled if one wants to use rollbacks.

PR #169, opened 2015-09-28
Escape methods

RomanSek:
Escape methods. They can be used to escape special characters in Solr queries. I'm not sure if this should be attached to this lib or It should be separate lib. What do you think?


acdha:
I believe @swistakm has been working on solrq for exactly this need:

https://pypi.python.org/pypi/solrq/

PR #170, opened 2015-10-04
Support for nested documents

skirsdeda:
This makes it possible to add nested documents (documents with children) via _Solr.add_. Includes tests as well, however {!child} query seems to be buggy in 4.10 so I left it disabled.

One thing worth discussing is the children key in doc dicts. I simply used "children" but maybe some safer option could be used (like '_children_' or '_childDocuments_' as in Solr's JSON request handler) so that it doesn't accidentally clash with simple fields.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -733,36 +733,44 @@ class Solr(object):
         self.log.debug("Found '%d' Term suggestions results.", sum(len(j) for i, j in res.items()))
         return res
 
-    def _build_doc(self, doc, boost=None, fieldUpdates=None):
-        doc_elem = ET.Element('doc')
+    def _build_doc(self, doc, boost=None, fieldUpdates=None, children=None):
+        def _one_doc(doc):
+            doc_elem = ET.Element('doc')
 
-        for key, value in doc.items():
-            if key == 'boost':
-                doc_elem.set('boost', force_unicode(value))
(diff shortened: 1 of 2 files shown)


skirsdeda:
Added NESTED_DOC_KEY='\_childDocuments\_' instead of hardcoded 'children' to match nested documents. Maybe this is a sensible default since it matches Solr's naming.
Also implementation was messed up previously since it only supported 1-level deep nesting.

So all is good now. Haystack support for nested documents on Solr and Elastic coming up :)

skirsdeda:
bump

skirsdeda:
@acdha bump.

ahankinson:
+1 Would be nice to have this in there.

ahankinson:
Can this be given a higher priority? I'm maintaining my own fork for this (https://github.com/ahankinson/pysolr/tree/fix-child-documents). I'd be happy to contribute it as a new pull request, but support for the `_childDocuments_` key would be nice.

acdha:
@ahankinson Since @skirsdeda sent a nice clean patch, it wasn't hard to update against the current master. The tests are running now: https://travis-ci.org/acdha/pysolr/builds/140901856

ahankinson:
fantastic, thank you very much.

PR #171, opened 2015-10-08
implement Solr.system_info

skirsdeda:
This is quite handy for checking solr version and such.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -962,6 +962,24 @@ class Solr(object):
 
         return data
 
+    def system_info(self):
+        """
+        Returns system info as returned by Solr's SystemInfoHandler
+        parsed into a dict
+        """
+        params = {'wt': 'json'}
+        params_encoded = safe_urlencode(params, True)
+
+        path = 'admin/system/?%s' % params_encoded
+        resp = self._send_request('get', path)
+        try:
+            data = json.loads(resp)
+        except ValueError as err:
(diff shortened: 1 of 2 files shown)


skirsdeda:
Also fixed SolrCoreAdmin. It was putting all params into request body for whatever reason and Solr treated all requests coming from SolrCoreAdmin as action=STATUS (default) requests. So everything other that SolrCoreAdmin.status didn't work at all (double checked this on Solr 4 and 5).

This also shows that tests/admin.py is useless since it only checks if solr returns good status code somewhere in response. But then SolrCoreAdmin should at lease use the same error checking code which currently is in Solr class. And I'm reluctant to do this refactoring until PRs #170 and #171 are merged.

skirsdeda:
@acdha This one fixes SolrCoreAdmin and adds ```system_info``` to get info about Solr instance. Should be merged.

RabidCicada:
Could we get some attention on system_info?  I need something like this for my solr6 upgrades on django_haystack to switch retrieval of suggestions @acdha

acdha:
Looks like we should start by rebasing this branch against the current master & checking our test coverage

RabidCicada:
I've worked around it by simply checking for type of suggestions object returned...but think an explicit solr check would be good.  It'll prevent future dead code hanging around with ???why do we check type here??? thoughts.  Instead the check will be explicit and dead/legacy code can be eliminated:).

PR #180, opened 2016-02-02
logging: use better logging messages on HTTP status codes != 200

swistakm:
We already extract error message from Solr responses and that is great. Unfortunately it can contain the data that may change with every request (like document id).

This creates an issue when user uses Sentry or other solution that captures logging or exceptions. Previous implementation causes many duplicated events in Sentry if message extracted using `self._extract_error(resp)` contained such variable data.

This change uses 'non-mutable' message that is complemented with extracted data that using string formatting option supplied by Python logging. Thanks to this, Sentry and other solutions can perform better grouping of loging messages (by status code).

This is approach that is already used in handling other errors.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -386,10 +386,12 @@ class Solr(object):
                       url, method, log_body[:10], end_time - start_time)
 
         if int(resp.status_code) != 200:
-            error_message = self._extract_error(resp)
-            self.log.error(error_message, extra={'data': {'headers': resp.headers,
-                                                          'response': resp.content}})
-            raise SolrError(error_message)
+            error_message = "Solr responded with an error (HTTP %s): %s"
+            solr_message = self._extract_error(resp)
+            self.log.error(error_message, resp.status_code, solr_message,
+                           extra={'data': {'headers': resp.headers,
+                                           'response': resp.content}})
+            raise SolrError(error_message % (resp.status_code, solr_message))
 
         return force_unicode(resp.content)
 


pysolr.py
```diff
@@ -386,10 +386,12 @@ def _send_request(self, method, path='', body=None, headers=None, files=None):
                       url, method, log_body[:10], end_time - start_time)
 
         if int(resp.status_code) != 200:
-            error_message = self._extract_error(resp)
-            self.log.error(error_message, extra={'data': {'headers': resp.headers,
-                                                          'response': resp.content}})
-            raise SolrError(error_message)
+            error_message = "Solr responded with error (HTTP %s): %s"
+            params = (resp.status_code, self._extract_error(resp))
```
acdha: Since this has only two values, maybe we should just pass them to log.error directly without using `*args` notation for clarity?

acdha:
I left one minor comment but otherwise :+1:

pysolr.py
```diff
@@ -386,10 +386,12 @@ def _send_request(self, method, path='', body=None, headers=None, files=None):
                       url, method, log_body[:10], end_time - start_time)
 
         if int(resp.status_code) != 200:
-            error_message = self._extract_error(resp)
-            self.log.error(error_message, extra={'data': {'headers': resp.headers,
-                                                          'response': resp.content}})
-            raise SolrError(error_message)
+            error_message = "Solr responded with error (HTTP %s): %s"
```
acdha: I'm not sure that it's worth a commit but do you think this would read better as “Solr responded with an error”?

swistakm: Good point. I will fix that.

pysolr.py
swistakm: No problem. Just tell me what do you prefer.

This:
```python
error_message = "Solr responded with error (HTTP %s): %s"
params = (resp.status_code, self._extract_error(resp))
self.log.error(error_message, params[0], params[1],
               extra={'data': {'headers': resp.headers,
                               'response': resp.content}})
raise SolrError(error_message % params)
```
or this:
```python
error_message = "Solr responded with error (HTTP %s): %s"
solr_message = self._extract_error(resp)
self.log.error(error_message, resp.status_code, solr_message,
               extra={'data': {'headers': resp.headers,
                               'response': resp.content}})
raise SolrError(error_message % (resp.status_code, solr_message)
```

acdha: I'd go with the second style since neither is particularly long

swistakm:
I have updated according to comments and squashed the fix branch to keep git log concise.

acdha:
Merged in a65399f65048100a7a6812cf74b98d7efe03718c

acdha:
I'm getting ready to tag v3.4.0

Issue #181, opened 2016-02-11
Solr.add supports commitWithin but not overwrite


robinsonkwame:
I have a use case for setting the overwrite attribute of the add tag to False.

Currently only the commitWithin attribute is supported. This could be done very simply by including overwrite as a Solr.add parameter and setting the message with its value if defined.

acdha:
If you want to send a pull request, I can review it.

robinsonkwame:
Thanks. See: https://github.com/toastdriven/pysolr/pull/182

PR #182, opened 2016-02-15
Solr.add supports commitWithin but not overwrite #181

robinsonkwame:
Adds overwrite=True|False support option to pysolr:Solr.add
Adds testing for Solr.add overwrite option

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -806,7 +806,7 @@ class Solr(object):
 
         return doc_elem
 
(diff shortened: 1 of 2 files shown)


acdha:
I have this parked in https://github.com/acdha/pysolr/tree/182-overwrite-support until someone has time to figure out why the tests are failing. At first I thought it was just Solr expecting the `overwrite` flag to be in the query string like e.g. `commit` but the tests fail either way.

robinsonkwame:
odd. I'll take a look at this week when I get the chance, thanks.

robinsonkwame:
Added id check in search, resolves test failure as far as I can see.

PR #184, opened 2016-02-19
Comments on update_function are not correct.

atuljangra:
I was going through the entire code because I wanted to fix an issue when the empty strings are not indexed in solr. While doing that, this thing caught my eye. The comment says "Passing `sanitize` as False ...", but there is no 'sanitize' parameter. I've updated that so that this reduces the confusion.


acdha:
Thanks!

PR #187, opened 2016-03-10
Add support for state.json SolrCloud, upgrade to 5.5 and add failover tests

upayavira:
This PR covers three things:
- Upgrade to Solr 5.5
- Extend clusterstate code to use the new state.json format used by Solr 5.x (should still be 4.x compatible)
- Add failover tests (should prove that failover code is working fine)

This code is not quite complete. Additional work required:
- I'm now getting a single failure on the SolrCloudTestCase.test_failover test
- According to a thread on the Lucene-dev list, I need to pay attention to live-nodes. That is
  once I've selected a node, confirm that it is up. I'm hoping this'll fix the above test
- Once everything has completed, and Solr has been stopped, the run-tests.py process just hangs. I haven't yet worked out why.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -1244,13 +1244,12 @@ class ZooKeeper(object):
 
     def watchCollection(self, collection):
         path = "/collections/%s/state.json" % collection
-        def watch(event):
-            data = self.zk.get(path)
+        def watch(event=None):
+            data = self.zk.get(path, watch=watch)
             self.collections[collection] = json.loads(data[0].decode("utf8"))[collection]
 
         try:
-            data = self.zk.get(path, watch=watch)
-            self.collections[collection] = json.loads(data[0].decode("utf-8"))[collection]
+            watch()
         except NoNodeError, e:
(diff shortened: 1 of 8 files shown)


acdha:
In my experience with this batch of code, the run-tests.py hangs are due to the questionable way Kazoo uses a bunch of async workers and has an atexit handler to cleanup which simply waits forever for the ZooKeeper server to return, which never happens since we've already shut down the test server at that point. That was what I was trying to address with commits like https://github.com/upayavira/pysolr/commit/e9d0f34fb4b66046667a407c3fa68f1deab6bae3 by explicitly closing connections before we shut down the server.

acdha:
I just made df2b3f3741c4b53fb467c32cacb14bd477c91663 to expose the max_retries parameter but it doesn't appear to completely avoid KazooClient blocking shutdowns.

I'm using b50b9a4b8427e6147636c223c3e1ab58f46de19e to see where in the Kazoo code the remaining endless retry is happening.

Issue #197, opened 2016-05-28
Issue with date


savalou:
# I have
* [x] Tested with the latest release
* [ ] Tested with the current master branch
* [ ] Searched for similar existing issues

## Actual behavior
 Trying to index some data in django, when I run ./manage.py build_index, I get the following:
```
File "/home/vagrant/django/local/lib/python2.7/site-packages/pysolr.py", line 386, in _send_request
    raise SolrError(error_message % (resp.status_code, solr_message))
SolrError: Solr responded with an error (HTTP 400): [Reason: Invalid Date in Date Math String:'2016-05-26T16:43:04.644145+00:00Z']
```

## Steps to reproduce the behaviour

1.

## Configuration

* Search engine version: Solr 6.0.0
* Python version:Python 2.7.6
* pysolr version: pysolr-3.5.0

pembo13:
Running into the very same issue.

The problem is a naive datetime yields a different format than a timezone aware version:

```python
import pytz
from datetime import datetime
datetime.utcnow().isoformat()
datetime.utcnow().replace(tzinfo = pytz.utc).isoformat()
```
yields

```
'2016-06-16T18:44:24.746108'
'2016-06-16T18:44:26.169965+00:00'
```

The problem seems to be in `pysolr`, in it's `_from_python` method.

Changing

```python
value = "%sZ" % value.isoformat()
```
to
```python
value = "%sZ" % ( value.strftime('%Y-%m-%dT%H:%M:%S.%f') if value.tzinfo else value.isoformat() )
```
fixes the problem, but I'm not sure that that's the best way to do it.

pembo13:
Can someone take a look at Pull #198

PR #201, opened 2016-07-26
Don't append 'Z' to timezone-aware dates

akuchling:
Here's an alternative fix for issue #197 that's a bit more wordy than pull request #198.  This version of the patch has some logic to figure out the timezone suffix to use, and also adds two tests that exercise the timezone-aware case.

Tests still pass locally.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -54,6 +54,8 @@ Fix
 
   `__init__` was not actually passing this to the ZooKeeper client
 
+- Don't append 'Z' to timezone-aware dates. Closes #197.  [Andrew Kuchling]
+
 Other
 ~~~~~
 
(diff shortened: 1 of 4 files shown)


akuchling:
It looks like the Travis CI failure is because the testing scripts no longer manage to download Solr.

acdha:
Yeah, we'll need the same fix as https://github.com/django-haystack/django-haystack/commit/89ef54a556e48f4597292eaf23f942284208785a to deal with the Apache mirrors removing the 4.x branch. I started https://github.com/acdha/pysolr/tree/dockerized-test-solr before going on vacation last week to start pivoting the tests over to use Docker so we can stop maintaining that bit of code.

akuchling:
I reworked the patch to only ever produce 'Z', because that's the only format that Solr4 and Solr6 want to accept.  I also applied the changes for the Apache mirrors, which results in most of the tests passing except for the ones running on PyPy.

acdha:
The PyPy tests are failing because Travis CI has an old PyPy version which cannot be used to enable SNI for HTTPS connections. It's annoying but clearly unrelated to this work.

PR #211, opened 2016-12-01
Allow fieldupdate 'set' to delete fields from a solr document

jogilder:
# Hey, thanks for contributing to pysolr. Please confirm that [the tests pass](https://github.com/django-haystack/pysolr/blob/master/README.rst#running-tests) locally

# Once your pull request has been submitted, the full test suite will be executed on https://travis-ci.org/django-haystack/pysolr/pull_requests. Pull requests with passing tests are far more likely to be reviewed and merged.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,7 +30,7 @@ Requirements
 ============
 
 * Python 2.7 - 3.5
-* Requests 2.0+
+* Requests 2.9.1+
 * **Optional** - ``simplejson``
 * **Optional** - ``kazoo`` for SolrCloud mode
 


pysolr.py
```diff
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+#this is copied from https://github.com/django-haystack/pysolr
```
acdha: This comment looks like it was intended for internal usage?

pysolr.py
```diff
@@ -9,7 +13,6 @@
 import re
 import time
 from xml.etree import ElementTree
-from pkg_resources import get_distribution, DistributionNotFound
```
acdha: This appears to be an inadvertent regression – it would probably be easiest to rebase your branch against the current master & re-push it to update the pull request.

pysolr.py
```diff
@@ -137,10 +137,8 @@ def force_bytes(value):
 def unescape_html(text):
     """
     Removes HTML or XML character references and entities from a text string.
-
```
acdha: It looks like there were a bunch of whitespace changes like this; I'd prefer to keep the commit as small as possible by omitting this sort of thing.

PR #212, opened 2016-12-02
updated build_doc so that an empty string or empty list can be passed…

jogilder:
# Hey, thanks for contributing to pysolr. Please confirm that [the tests pass](https://github.com/django-haystack/pysolr/blob/master/README.rst#running-tests) locally

# Once your pull request has been submitted, the full test suite will be executed on https://travis-ci.org/django-haystack/pysolr/pull_requests. Pull requests with passing tests are far more likely to be reviewed and merged.
… to solr if fieldupdates is used. An extra attribute of null is added to the message. This allows a an element to be removed from the solr document. Previously, the value was ignored and no change was made.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -798,6 +798,28 @@ class Solr(object):
         self.log.debug("Found '%d' Term suggestions results.", sum(len(j) for i, j in res.items()))
         return res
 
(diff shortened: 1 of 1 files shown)


jogilder:
so is this fixed then?

PR #220, opened 2017-05-27
Enhancements to the README

alexwlchan:
* Clarify that pysolr can be installed with pip
* Now that Python 3.6 is in the build matrix (788ca8b), add Python 3.6 to the list of supported Pythons

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,9 +36,14 @@ Requirements
 Installation
 ============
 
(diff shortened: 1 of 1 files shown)


acdha:
Thanks - that bit about running setup.py is at least 8 years old so it was definitely due for modernization:

https://github.com/django-haystack/pysolr/commit/d09af73edeaa86fed3fcf842fb887d0660344a79

PR #221, opened 2017-07-17
Add ability to declare Solr object with default commit policy

efagerberg:
Since solr has hooks to auto commit after X documents or X seconds, it is best that the `Solr` methods `get`, `delete`, and `_update` do not have a policy of committing by default.

One may already assume this to be false by default since it is good practice to limit commits. But incase users rely on this functionality, there is the ability to declare the default policy in the `Solr` constructor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@
 solr*.tgz
 solr-app
 solr
-*.pyc
-logs
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 solr*.tgz
 solr-app
 solr
+*.pyc
+logs
--- a/README.rst
+++ b/README.rst
@@ -200,9 +200,9 @@ Custom Commit Policy
 
     # Setup a Solr instance. The trailing slash is optional.
     # All request to solr will result in a commit
-    solr = pysolr.Solr('http://localhost:8983/solr/core_0/', search_handler='/autocomplete', commit_by_default=True)
+    solr = pysolr.Solr('http://localhost:8983/solr/core_0/', search_handler='/autocomplete', always_commit=True)
 
-``commit_by_default`` signals to the Solr object to either commit or not commit by default for any solr request.
+``always_commit`` signals to the Solr object to either commit or not commit by default for any solr request.
 Be sure to change this to True if you are upgrading from a version where the default policy was alway commit by default.
 
 Functions like ``add`` and ``delete`` also still provide a way to override the default by passing the ``commit`` kwarg.
--- a/README.rst
+++ b/README.rst
@@ -193,6 +193,24 @@ If your Solr servers run off https
 	solr = pysolr.SolrCloud(zookeeper, "collection", verify=path/to/cert.perm)
 
 
+Custom Commit Policy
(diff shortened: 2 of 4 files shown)


efagerberg:
@acdha you mind giving this a look over?

.gitignore
```diff
@@ -1,4 +1,6 @@
 .tox
 solr*.tgz
 solr-app
-solr
```
acdha: This change is unrelated

pysolr.py
```diff
@@ -898,11 +905,12 @@ def add(self, docs, boost=None, fieldUpdates=None, commit=True, softCommit=False
         m = force_unicode(m)
 
         end_time = time.time()
+        commit = self._get_commit(commit)
```
acdha: Do we really need to make this call here when it's just calling `_update()` which does the same check? I'm thinking we could remove it in both `_add` & `_delete` and inline the `_get_commit` check into `_update`.

efagerberg:
Thanks I will review this sometime tomorrow.

acdha:
Thanks for sending the request - it's a good idea but this has been a busy period at work for me to get around to OSS reviews.

efagerberg:
@acdha you were totally right about `_get_commit`. Fixed that up and added a those tests.

acdha:
Can you rebase this against the current master branch & confirm that the tests still pass?

README.rst
```diff
@@ -156,6 +156,25 @@ with: ``mlt``, ``update``, ``terms`` etc. The specific methods of pysolr's ``Sol
...
 
 
+Custom Commit Policy
+~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    # Setup a Solr instance. The trailing slash is optional.
+    # All request to solr will result in a commit
+    solr = pysolr.Solr('http://localhost:8983/solr/core_0/', search_handler='/autocomplete', commit_by_default=True)
```
mlissner: Maybe this should be something like send_commits or something? by_default doesn't sound right to me.

efagerberg:
@acdha sorry about the wait on this, rebased onto master. @mlissner I have changed the parameter name to `always_commit` is that more appropriate?

mlissner:
And this should not be closed either. Can't say I'm a fan of the stale bot.

acdha:
Thanks for updating it. It looks good to me. The only question I had is whether we should default to `always_commit=True` for backwards-compatibility on the assumption that people who aren't pressed on are less likely to read the documentation but I don't feel especially strongly about it.

mlissner:
I think we concluded that switching to `always_commit=False` was better because it:

 - would be a good thing for most people, who don't realize that commits should be infrequent in Solr.
 - would be a good reason to do a major version bump
 - should have minimal impact because of Solr's default commit policy (which regularly does commits)

mlissner:
Oh, and I do feel pretty strongly about this, because always_commit=True is a weird default that'd definitely mess some people up.

acdha:
Fine by me

efagerberg:
Thanks folks 😄

acdha:
@efagerberg if you want to draft that wording in a pull request it's a good idea.

PR #227, opened 2017-11-10
Test pysolr against multiple versions of Solr, fixes #223

### LINKED ISSUE #223: Test suite should run for multiple solr versions
Currently pysolr's test suite only runs on Solr 4.10.4

According to http://django-haystack.readthedocs.io/en/master/backend_support.html pysolr supports 3.5+ but the test suite does not reflect that initiative.

The tests should be able to run against many versions of Solr in order to assert that pysolr works with those versions.

jnovinger:
This PR adds test runs against Solr 5.x, 6.x, and 7.x for a standard, stand-alone Solr configuration as mentioned in #223 by @efagerberg  . This PR adds a docker-compose.yml, Dockerfile, and Solr configuration skeleton for testing Pysolr against Solr 5.x, 6.x, and 7.x. Testing against Solr 4.x is unchanged and still implemented using the start-solr-test-server.sh script and a local download of Solr 4.

Due to time constraints, this does not include tests for Pysolr's SolrCloud support for Solr 5.x+. These tests have been marked as "failure allowed" in the Travis configuration, so that they don't break build status in the short-term.

This PR includes 2 commits, one that represents the bulk of the changes (including a skeleton Solr config) and one that removes much of the bloat (e.g. comments) from the Solr config as is reasonable. If you'd prefer, I can remove the second commit, but chose to add it to reduce the overall size of the PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,57 @@
-sudo: false
+sudo: required
 language: python
 python:
-    - "2.7"
-    - "3.3"
-    - "3.4"
-    - "3.5"
-    - "3.6"
-    - "pypy"
+  - "2.7"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+  - "3.6"
+  - "pypy"
 
 cache:
   apt: true
   pip: true
   directories:
-      - $HOME/download-cache
+    - $HOME/download-cache
 
 env:
-    - SOLRCLOUD=false
-    - SOLRCLOUD=true
+  - SOLR_VERSION="4.10.4" SOLRCLOUD=false
+  - SOLR_VERSION="4.10.4" SOLRCLOUD=true
+  - SOLR_VERSION="5.5" SOLRCLOUD=false
+  - SOLR_VERSION="5.5" SOLRCLOUD=true
+  - SOLR_VERSION="6.6" SOLRCLOUD=false
+  - SOLR_VERSION="6.6" SOLRCLOUD=true
+  - SOLR_VERSION="7.1" SOLRCLOUD=false
+  - SOLR_VERSION="7.1" SOLRCLOUD=true
 
 matrix:
-    allow_failures:
-        - python: 'pypy'
+  allow_failures:
+    - env: SOLR_VERSION="5.5" SOLRCLOUD=true
+    - env: SOLR_VERSION="6.6" SOLRCLOUD=true
+    - env: SOLR_VERSION="7.1" SOLRCLOUD=false
+    - env: SOLR_VERSION="7.1" SOLRCLOUD=true
+    - python: 'pypy'
+
+services:
+  - docker
 
 addons:
   apt_packages:
     - default-jdk
 
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install docker-ce
+
 install:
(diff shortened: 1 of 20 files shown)


jnovinger:
@acdha I am unclear on how the Apache license statements, present in the Solr config files, plays with Pysolr's choice of BSD license. If you have guidance on how to handle that, please let me know.

jnovinger:
I forgot to mention that there are some failing tests when Pysolr is pointed at a Solr 7.x instance. This appears to be because Solr 7.x+ returns JSON from admin endpoints, rather than XML. I looked at the failures briefly and was unable to tell from the responses if the actions actually completed successfully (e.g. a core was created). I marked all Solr 7.x tests as "failure allowed" as well, thinking that this may be better addressed later.

acdha:
@jnovinger can we support older versions of Solr by adding wt=json to the admin URLs? (It's been a long time since I've looked at those) I think by now it'd be reasonable to say that e.g. we only support Solr 5 or 6+ going forward since people with older Solr installs can use the current releases.

jnovinger:
@acdha good questions, I'll take a look and let you know as soon as I can.

jnovinger:
Gah, please don't close this @stalebot. I've been thinking about this recently and meaning to revisit it.

Andrew-Chen-Wang (commented):
Hey @jnovinger @acdha pinging just to see if this can be ramped up again. I'm working in a Python 3.8 environment and hopefully about to begin Solr usage with their somewhat latest stable version. Would be great to see multi-version Solr testing (and if unit testing can include Python 3.8 as a supported version, which shouldn't be an issue to me since it was a lot of performance gains from Python changelog)!

Just my two cents: when I'm running something like Elasticsearch in Travis, I'll just install it directly without using Docker. That might be able to cut down on some files. Not sure though so I've never used Solr yet.

cclauss:
Please rebase.

jnovinger:
I'll see if I can find some time to get this rebased soon and updated.

I haven't used Solr in a while, so I might be a little rusty.

cclauss:
@jnovinger Please rebase or close.

jnovinger:
@cclauss apologies for letting this get stale ... again. I've rebased it and will be able to take a look at getting it running correctly agains SOON:tm:. I have some PTO coming up at the end of this week and will dig in then.

cclauss:
Cool.  To placate `codespell`, modify the last line of `pyproject.toml` from `skip="./.*"` to `skip="./.*,*.xml"`

.travis.yml
```diff
@@ -0,0 +1,57 @@
+sudo: required
```
cclauss, line 1: Travis CI is no longer free to open source projects so we use GitHub Actions instead.

jnovinger: Yep, figured that would be where I need to do most of the work.

Honestly, I thought I told git to remove that file during the rebase. But, perhaps it's good to have for reference until the PR is ready for re-review. It will definitely be gone by then.

Andrew-Chen-Wang (approved):
Hi. I haven't touched Solr in awhile though, but happy to unblock as it's just tests

PR #229, opened 2017-12-24
Convert python sets to Solr supported format

samj1912:
I recently migrated to pysolr from solrpy and noticed it didn't support sets. All it needed was a simple isinstance check, so well here we are.


samj1912:
@acdha  can you review this please?

acdha:
Thanks - I missed that over the holidays

PR #230, opened 2018-01-08
Termvectors

biggosh:
Support for termvectors added


AlJohri:
related: https://github.com/django-haystack/pysolr/pull/121

Issue #130, opened 2014-09-17
Error in "extract" method, if spaces or special characters in the filename of the provided file object


carina21:
Hi all,

pysolr is a good library to communicate between python and solr server! But I found a little bug in "extract" method:
If there exists spaces or special characters in the filename of the provided file object, then occurs following error:
pysolr        ERROR   [Reason: None]
{"responseHeader":{"status":400,"QTime":1},"error":{"msg":"missing content stream","code":400}}
pysolr        ERROR   Failed to extract document metadata: [Reason: None]
{"responseHeader":{"status":400,"QTime":11},"error":{"msg":"missing content stream","code":400}}
Traceback (most recent call last):
  File "pysolr.py", line 905, in extract
    files={'file': (file_obj.name, file_obj)})
  File "/home/cw/Aptana Studio 3 Workspace/arcdoc/pysolr.py", line 321, in _send_request
    raise SolrError(error_message)
SolrError: [Reason: None]
{"responseHeader":{"status":400,"QTime":11},"error":{"msg":"missing content stream","code":400}}

To fix this bug, I added following code lines:
in line 891: filename = urllib.quote(file_obj.name)

in line 897: files={'file': (filename, file_obj)})

Here an overview of the extract method with the fixed bug:
    def extract(self, file_obj, extractOnly=True, **kwargs):
        """
        POSTs a file to the Solr ExtractingRequestHandler so rich content can
        be processed using Apache Tika. See the Solr wiki for details:

```
        http://wiki.apache.org/solr/ExtractingRequestHandler

    The ExtractingRequestHandler has a very simply model: it extracts
    contents and metadata from the uploaded file and inserts it directly
    into the index. This is rarely useful as it allows no way to store
    additional data or otherwise customize the record. Instead, by default
    we'll use the extract-only mode to extract the data without indexing it
    so the caller has the opportunity to process it as appropriate; call
    with ``extractOnly=False`` if you want to insert with no additional
    processing.

    Returns None if metadata cannot be extracted; otherwise returns a
    dictionary containing at least two keys:

        :contents:
                    Extracted full-text content, if applicable
        :metadata:
                    key:value pairs of text strings
    """
    if not hasattr(file_obj, "name"):
        raise ValueError("extract() requires file-like objects which have a defined name property")

    params = {
        "extractOnly": "true" if extractOnly else "false",
        "lowernames": "true",
        "wt": "json",
    }
    params.update(kwargs)
    filename = urllib.quote(file_obj.name)
[...]

PR #234, opened 2018-03-14
pysolr.extract() handle spaces and special characters in filenames (with added passing test cases)

dwvisser:
I created a test case slightly modified from test_extract in test_client.py. See Issue #130 
Thanks to @carina21 for the suggested fix.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -1034,7 +1034,7 @@ class Solr(object):
             "wt": "json",
         }
         params.update(kwargs)
-        filename = quote(file_obj.name)
+        filename = quote(file_obj.name.encode('utf-8'))
         try:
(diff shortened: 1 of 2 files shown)


dwvisser:
I didn't run the tests under Python 2.7 initially, and thus got surprised by failures in Travis. I fixed by mimicking what Py3 `urllib.parse.quote` does when it encounters Unicode strings: encodes to bytes using UTF8.

dwvisser:
I notice this PR isn't mentioned in the v3.8.1 release notes.

acdha:
Ah, yeah, I just updated https://github.com/django-haystack/pysolr/releases/tag/v3.8.1

PR #241, opened 2018-09-06
Adding ChangeLog note about default commit policy

efagerberg:


efagerberg:
@acdha here is my draft for the change log addition about the commit policy.

PR #252, opened 2019-01-29
Issue 251: Iterate through a cursor

ke4roh:
Iterating over a Result will give all its documents and query for more if and only if there is cursorMark (and sort order) defined in the original query.

One question: What about if someone wants to use a custom Results object AND a cursorMark? I'd be interested in your recommendations on whether or how to handle that.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -240,6 +240,9 @@ class Results(object):
     Individual documents can be retrieved either through ``docs`` attribute
     or by iterating over results instance.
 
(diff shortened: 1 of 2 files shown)


ke4roh:
ping

ke4roh:
Thoughts on this? In my experience, it makes cursors much easier.  What other angles do you see?

ke4roh:
This PR is superseded by pull request #276 which resolves the merge conflicts.

PR #254, opened 2019-02-14
Enable sonarcloud for Travis builds

acdha:
This should be enough configuration to run SonarQube after each build


acdha:
Completed in #258

PR #256, opened 2019-03-10
Add JSON interface to increase performance

trieder83:
Add JSON interface to increase performance by 15x to 20x and reduce CPU load significantly.
I noticed that for larger loads pysolr is cpu bound, which is mainly caused from the xml transformation. Using the JSON Array API removes the need for the XML transformation.
On my machines the load speed could be increased about 15x to 20x times, depending on the documents structure.


acdha:
Thanks for working on this - that should be a welcome boost for many users. I wanted to check how far back JSON support goes - it seems like this might also be a good time to drop some of that gnarled code handling errors from really old versions of Solr

trieder83:
The oldest documentation I could find on the official solr homepage is from Solr 6.6. That version includes the JSON and JSON Array Interface already. (https://lucene.apache.org/solr/guide/6_6/uploading-data-with-index-handlers.html)

Earlier Versions as 4.0 mention a JSON API, but I could not find a proper documentation so far. 
For sure it works with 6.6+ (6 June 2017)

From http://lucene.apache.org/solr/news.html
  Solr 4.8.0: JSON updates now support nested child documents

acdha:
I rebased & merged this: https://github.com/django-haystack/pysolr/commit/a6664ca25247e753f1ab3d2196b96094f4c32731

PR #257, opened 2019-04-24
Improve performance of xml character filtering

jamespeacock:
Regexp can filter out invalid bytes from XML strings faster than `join()`


pysolr.py
```diff
@@ -225,7 +225,7 @@ def clean_xml_string(s):
 
     http://stackoverflow.com/questions/8733233/filtering-out-certain-bytes-in-python
     """
-    return ''.join(c for c in s if is_valid_xml_char_ordinal(ord(c)))
+    return re.sub(u'[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD\U00010000-\U0010FFFF]+', '', s)  #''.join(c for c in s if is_valid_xml_char_ordinal(ord(c)))
```
acdha: I'd delete the code in the comment at the end of the line.

Since this is used a lot it'd probably be best to compile it once (e.g. `VALID_XML_CHARACTERS = re.compile(…)` outside of that function) rather than relying on the regex cache.

Since `is_valid_xml_char_ordinal` isn't used anywhere else, it can be deleted.

jamespeacock: Sorry for getting around to this so late. I have made requested changes.

acdha:
We should probably get some before/after benchmarks for this, too.

cclauss:
Please rebase.

PR #263, opened 2019-06-17
Fix for SolrCloud unknown collection issue

### LINKED ISSUE #189: Error: Unknown collection - SolrCloud __init__
Hi!
I'm having an error when trying to create a new SolrCloud object.

```
File "/home/patodb/Repos/hdfs_solr_indexer/indexer.py", line 23, in __init__
    self.solr = SolrCloud(zk, collection)
File "/home/patodb/Repos/hdfs_solr_indexer/pysolr.py", line 1148, in __init__
    url = zookeeper.getRandomURL(collection)
File "/home/patodb/Repos/hdfs_solr_indexer/pysolr.py", line 1289, in getRandomURL
    return random.choice(self.getHosts(collname, only_leader=False)) + "/" + collname
File "/home/patodb/Repos/hdfs_solr_indexer/pysolr.py", line 1255, in getHosts
    raise SolrError("Unknown collection: %s", collname)
pysolr.SolrError: (u'Unknown collection: %s', 'boletines_cba')
```

The error is raised in the `getHosts` method of the `ZooKeeper` class. Debugging the program I found that `self.collections` and `self.aliases` are not being initialized correctly when creating the ZooKeeper object. **Both variables have the default value {}** at the time `getRandomURL()` is executed and I don't see a workflow where these values are overriden.

The Zookeeper object (from Kazoo) seems fine, it resolves both the urls and chroot correctly. The server is running and the collection exists. Any ideas about why is this happening?

Regards,

csvital:
Solution from Qinwenchao's answer

Fixes #189

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -1328,7 +1328,6 @@ class ZooKeeper(object):
                 LOG.warning("No cluster state available: no collections defined?")
             else:
                 self.collections = json.loads(data.decode('utf-8'))
-                self.collections.update(json.loads(data.decode('utf-8')))
                 LOG.info('Updated collections: %s', self.collections)
 
         @self.zk.ChildrenWatch(ZooKeeper.LIVE_NODES_ZKNODE)
--- a/pysolr.py
+++ b/pysolr.py
@@ -1284,6 +1284,8 @@ class ZooKeeper(object):
     LIVE_NODES_ZKNODE = '/live_nodes'
     ALIASES = '/aliases.json'
     CLUSTER_STATE = '/clusterstate.json'
+    COLLECTION_STATUS = '/collections'
+    COLLECTION_STATE = '/collections/%s/state.json'
     SHARDS = 'shards'
     REPLICAS = 'replicas'
     STATE = 'state'
@@ -1326,6 +1328,7 @@ class ZooKeeper(object):
                 LOG.warning("No cluster state available: no collections defined?")
             else:
                 self.collections = json.loads(data.decode('utf-8'))
+                self.collections.update(json.loads(data.decode('utf-8')))
                 LOG.info('Updated collections: %s', self.collections)
 
         @self.zk.ChildrenWatch(ZooKeeper.LIVE_NODES_ZKNODE)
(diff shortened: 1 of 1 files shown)


pysolr.py
```diff
@@ -1326,6 +1328,7 @@ def watchClusterState(data, *args, **kwargs):
                 LOG.warning("No cluster state available: no collections defined?")
             else:
                 self.collections = json.loads(data.decode('utf-8'))
+                self.collections.update(json.loads(data.decode('utf-8')))
```
acdha: Doesn't this have the same effect as the previous line?

acdha:
https://github.com/django-haystack/pysolr/commit/17e9e7739da6f4d6798f8c9a774da8df9c2ff4b7 should fix that invalid URL error. Feel free to rebase against master.

csvital:
> [17e9e77](https://github.com/django-haystack/pysolr/commit/17e9e7739da6f4d6798f8c9a774da8df9c2ff4b7) should fix that invalid URL error. Feel free to rebase against master.



> [17e9e77](https://github.com/django-haystack/pysolr/commit/17e9e7739da6f4d6798f8c9a774da8df9c2ff4b7) should fix that invalid URL error. Feel free to rebase against master.

I've fixed this

pysolr.py
```diff
@@ -1346,6 +1349,19 @@ def watchAliases(data, stat):
...
 
+        def watchClusterDetailState(data, *args, **kwargs):
+            if not data:
+                LOG.warning("No cluster state available: no collections defined?")
+            else:
+                self.collections.update(json.loads(data.decode('utf-8')))
+                LOG.info('Updated collections: %s', self.collections)
+
+        @self.zk.ChildrenWatch(ZooKeeper.COLLECTION_STATUS)
+        def watchClusterState(children):
```
acdha: This is what's causing the CI failure since it's the same name as the earlier `watchClusterState` method:

https://travis-ci.org/django-haystack/pysolr/jobs/546780831#L370

Since this isn't really the right name, maybe we should rename this?

```suggestion
        def watchCollectionStatus(children):
```

pysolr.py
```diff
@@ -1284,6 +1284,8 @@ class ZooKeeper(object):
     LIVE_NODES_ZKNODE = '/live_nodes'
     ALIASES = '/aliases.json'
     CLUSTER_STATE = '/clusterstate.json'
+    COLLECTION_STATUS = '/collections'
+    CLUSTER_DETAIL_STATE = '/collections/%s/state.json'
```
acdha: Is this the best name or should that be something like `COLLECTION_STATE`?

pysolr.py
```diff
@@ -1346,6 +1349,19 @@ def watchAliases(data, stat):
                 self.aliases = None
             LOG.info("Updated aliases: %s", self.aliases)
 
+        def watchClusterDetailState(data, *args, **kwargs):
```
acdha: I'm wondering whether this might be better named `watchCollectionState`

pysolr.py
csvital: You're right I followed Qinwenchao's answer but there was some naming issues, resolved

Issue #251, opened 2019-01-18
Iterate through a cursor


ke4roh:
# I have
* [X] Tested with the latest release
* [X] Tested with the current master branch
* [X] Searched for similar existing issues

## Expected behaviour
The Results object offers an iterator over a cursor of results.

## Actual behaviour
I have to write/run something like this code every time I want to iterate through a cursor:

```
def iter_records(solr, **qparams):
    result = solr.search(**qparams)

    while True:
        for doc in result.docs:
            yield doc
        if result.raw_response['responseHeader']['params'].get('cursorMark',None) == result.raw_response.get("nextCursorMark",None):
            break
        qparams['cursorMark']=result.raw_response["nextCursorMark"]
        result = solr.search(**qparams)
```
## Steps to reproduce the behaviour

```
solr = Solr(...)
result = solr.query(q="*:*",cursorMark="*",sort="id asc")
for doc in result:
    print(str(doc))
```
(This implies one particular implementation, which may not be the best, but it's a start.)

## Configuration

* Operating system version: any
* Search engine version: any
* Python version: 3.6
* pysolr version: 3.7.0

ke4roh:
I have a patch for this which passes a lambda function to the Results object to keep track of how to search (parameters and which Solr instance).

Regarding the iterator implementation, we don't want to break the existing behavior - iterating through results.docs should return the docs retrieved, and so forth.  A separate iterator is in order.  A list (or tuple) is not appropriate because we have no guarantee that the entire result set might fit in memory.  The __iter__() function seems to me like a reasonable enough place to put the feature.  (Will people find it?)

ke4roh:
Another thought I had about implementation is that we *could* mimic the behavior of an array for those who want it, with the primary use case being the iterator.  The drawback would be that it would take O(N) time to get to any element in the array, unless it happened to be in the current Result.  Caching could expedite many cases (caching neighboring results and tracking incices for a finite number of cursorMarks), but that seems like a lot of work for something that is fundamentally an iterator.

acdha:
Yeah, I'm generally in favor of not adding caching without some sort of clear user-visible indication that potentially non-trivial amounts of memory will be used. I think an iterator is a good fit for this as long as it has a reasonable chunk size so we don't hide a huge number of requests.

ke4roh:
[...]

PR #276, opened 2019-08-08
Issue #251: Iterate through a cursor

ke4roh:
Re-merge pull request #252 into the latest code.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -245,6 +245,9 @@ class Results(object):
     Individual documents can be retrieved either through ``docs`` attribute
     or by iterating over results instance.
 
+    Optional ``next_page_query`` argument is a callable to be invoked when
+    iterating over the documents from the result.
+
     Example::
 
         results = Results({
@@ -275,7 +278,7 @@ class Results(object):
     use with features which change the response format.
     """
 
-    def __init__(self, decoded):
+    def __init__(self, decoded, next_page_query=None):
         self.raw_response = decoded
 
         # main response part of decoded Solr response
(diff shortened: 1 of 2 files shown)


pysolr.py
```diff
@@ -820,7 +829,16 @@ def search(self, q, search_handler=None, **kwargs):
             # cover both cases: there is no response key or value is None
             (decoded.get("response", {}) or {}).get("numFound", 0),
         )
-        return self.results_cls(decoded)
+
+        if "cursorMark" in params and params["cursorMark"] != \
```
acdha: I'm wondering whether this would be cleaner as `cursorMark = params.get("cursorMark")`, etc. especially with the fallback logic on the `decoded.get()` call.

pysolr.py
```diff
@@ -820,7 +829,16 @@ def search(self, q, search_handler=None, **kwargs):
...
+                decoded.get('nextCursorMark', params["cursorMark"]):
+            def next_page_query():
+                nextParams = dict(params)
```
acdha: This might be more clear about the intention:
```suggestion
                nextParams = params.copy()
```

pysolr.py
```diff
@@ -820,7 +829,16 @@ def search(self, q, search_handler=None, **kwargs):
...
+                nextParams["cursorMark"] = decoded['nextCursorMark']
```
acdha:
```suggestion
                nextParams["cursorMark"] = decoded["nextCursorMark"]
```

pysolr.py
ke4roh: Maybe this is better?:
```
        cursorMark = params.get("cursorMark", None)
        if cursorMark != decoded.get("nextCursorMark", cursorMark):
```

acdha: I like that and it avoids an extra lookup

ke4roh: cool

ke4roh:
What else?

PR #278, opened 2019-08-10
run-tests: Avoid unnecessary UTF-8 output

### LINKED ISSUE #277: run-tests: script crashes with C locale due to unicode chars being printed
# I have
* [x] Tested with the latest release
* [x] Tested with the current master branch
* [x] Searched for similar existing issues

## Expected behaviour
Test script exits successfully.

## Actual behaviour
Test script crashes after tests succeed:
```
Ran 106 tests in 40.806s

OK
Traceback (most recent call last):
  File "run-tests.py", line 28, in <module>
    main()
  File "run-tests.py", line 23, in main
    print("Tests complete; halting Solr servers…")
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2026' in position 36: ordinal not in range(128)
```

## Steps to reproduce the behaviour

1. `export LC_ALL=C`
2. `python2.7 run-tests.py`

## Configuration

* Operating system version: Gentoo Linux
* Search engine version: apparently 4.10.4 gets fetched for tests
* Python version: 2.7.16
* pysolr version: git master

mgorny:
Replace the UTF-8 ellipsis with plain ASCII.  Printing it is currently
broken with non-UTF-8 locale, and there is really no technical reason to
add even more complexity for a status message, given that everything
else works just fine.

Fixes #277.

### DIFF
--- a/run-tests.py
+++ b/run-tests.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
-# encoding: utf-8
 
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import absolute_import, print_function
 
 import faulthandler
 import signal
@@ -20,7 +19,7 @@ def main():
     try:
         unittest.main(module="tests", verbosity=1)
     finally:
-        print("Tests complete; halting Solr servers…")
+        print("Tests complete; halting Solr servers...")
         test_utils.stop_solr()
 
 


cclauss:
Please rebase to run the tests.

mgorny:
I had to create a new PR because my fork has been deleted since. It's #395 now.

PR #356, opened 2021-01-22
Escape special characters in the XML generated in the delete() method

ramayer:
Type pysolr .delete() method was not escaping special characters like <, >, and & in the XML it was generating.

In particular attempts to delete documents by IDs like this: 

        self.solr.delete(id='cats<dogs')

would give error messages, and attempts to delete documents like this:

        self.solr.delete(q='id:*</query><query> id:999 AND id:9999')

could delete more than you would have expected.   This pull request (with test cases) escape the IDs and Lucene Query Expressions in the XML generation step.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -9,6 +9,7 @@ import random
 import re
 import time
 from xml.etree import ElementTree
+import xml.sax.saxutils
 
 import requests
 from pkg_resources import DistributionNotFound, get_distribution, parse_version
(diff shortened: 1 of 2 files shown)


ramayer:
Note that this has the same functionality as [#358].

This version took the approach of using minimal changes building XML the same way.

The other version took the approach of changing Delete to use ElementTree to build the XML instead of python string manipulation.

I not no preference which approach is approved -- they both generate the same XML.

ramayer:
Based on comments in [#357], the different implementation of this fix under pull request [#358]  is a better approach.

PR #358, opened 2021-01-22
Escape xml in delete() using ElementTree

ramayer:
A different approach to [#356] using ElementTree as suggested in a comment.

Type pysolr .delete() method was not escaping special characters like <, >, and & in the XML it was generating.

In particular attempts to delete documents by IDs like this: 

        self.solr.delete(id='cats<dogs')

would give error messages, and attempts to delete documents like this:

        self.solr.delete(q='id:*</query><query> id:999 AND id:9999')

could delete more than you would have expected.   This pull request (with test cases) escapes the IDs and Lucene Query Expressions in the XML generation step.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -9,7 +9,6 @@ import random
 import re
 import time
 from xml.etree import ElementTree
-import xml.sax.saxutils
 
 import requests
 from pkg_resources import DistributionNotFound, get_distribution, parse_version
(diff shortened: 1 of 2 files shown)


ramayer:
Note that this has the same functionality as [#356 ].

This version took the approach of using minimal changes building XML using ElementTree as the comments under [357] suggested  

The other version took the approach of making a more minimal change, building the XML using python strings like the delete() method did earlier.

I not no preference which approach is approved -- they both generate the same XML.

ramayer:
My initial pull request failed the flake8 tests with too long lines and improper whitespace.  I pushed an updated version this morning.

acdha (approved):

PR #359, opened 2021-01-22
Show an example of how to use cursor traversal

ke4roh:
I've looked up cursor traversal in the test enough times to figure it's worth putting it in the readme.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,6 +112,10 @@ Basic usage looks like:
         'hl.fragsize': 10,
     })
 
+    # Traverse a cursor using its iterator:
+    for doc in solr.search('*:*',fl='id',cursorMark='*'):
+        print(doc['id'])
+
     # You can also perform More Like This searches, if your Solr is configured
     # correctly.
     similar = solr.more_like_this(q='id:doc_2', mltfl='text')


acdha (approved):

acdha:
Thanks!

alishah730:
why its running in infine loop,
first cursor * it should stop 
when i'll hit again with nextcursor it will give next set of result 
right?

ke4roh:
I'm not sure.  I forgot to include sort order (which Solr requires) in the parameters, so you would get an error without that, but otherwise, you should be able to use the iterator over the result set and get ALL of the results.  Maybe you are getting more results than you expected?  Have a look at the [tests for this functionality](https://github.com/django-haystack/pysolr/blob/5bd07eaed3fab110c9c41e05a765f9d95ac731b2/tests/test_client.py#L283) and see if that covers your situation.

alishah730:
I guess its not working how pagination expected to work
results = solr.search(q='*:*', cursorMark="*", rows=2, sort="id asc")
this should return only two rows with netxCursorMark="879chwodfgw9eiofsiodcgufio"
and next search query will be 

results = solr.search(q='*:*', cursorMark="879chwodfgw9eiofsiodcgufio", rows=2, sort="id asc")
this will return next 2 records with netxCursorMark="ouji89vnwvebuvwoepojveniowviwonv"

results = solr.search(q='*:*', cursorMark="ouji89vnwvebuvwoepojveniowviwonv", rows=10, sort="id asc")
this will return next 10 records with netxCursorMark="9odoucwenu9hwe0ewijvw"

and so on.

but now if I have 1M records its. iterating all over which is not expected 

same explained here 

https://lucene.apache.org/solr/guide/8_6/pagination-of-results.html

ke4roh:
PySolr will operate the cursorMark and perform subsequent queries for you when you iterate on a result having a cursor mark, as per the feature we're discussing here.  If you want only the results from that singular trip to Solr, use result.docs.  

If that doesn't help with your question, could you try to write a failing test?

PR #374, opened 2021-10-01
Perform adds with field updates using JSON

mitchelljkotler:
This pull request adds support for doing field updates using JSON instead of XML.  I have a need for this due to the intersection of indexing nested documents and field updates.

Indexing nested documents can now work with labelled fields:
https://solr.apache.org/guide/8_0/indexing-nested-documents.html#xml-examples
https://solr.apache.org/guide/8_0/indexing-nested-documents.html#json-examples
And this is preferred over the anonymous way, using the special key \_childDocuments\_ which pysolr currently supports.

This currently works when using JSON serialization (the default), but the child documents end up being serialized as a string when using XML serialization (which happens when field updates is set).

Field updates can be done through JSON (the documentation actually doesn't even show the XML syntax anymore):
https://solr.apache.org/guide/8_0/updating-parts-of-documents.html#example-updating-part-of-a-document

I have implemented field updates with JSON serialization.  I did not add any tests, but I believe the existing field update tests validate the JSON implementation.

This leaves XML serialization to be used only for index time boosts, which have been removed in version 7:
https://solr.apache.org/docs/7_0_0/changes/Changes.html#v7.0.0.upgrading_from_solr_6.x
> 11. Index-time boosts are not supported anymore. If any boosts are provided, they will be ignored by the indexing chain. As a replacement, index-time scoring factors should be indexed in a separate field and combined with the query score using a function query. 

I believe this means the XML serialization can be completely dropped, but I wanted to check with you if this is something you would be willing to merge in.  If it is, I would be happy to continue working on this.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -916,9 +916,19 @@ class Solr(object):
         )
         return res
 
-    def _build_json_doc(self, doc):
-        cleaned_doc = {k: v for k, v in doc.items() if not self._is_null_value(v)}
+    def _build_json_doc(self, doc, fieldUpdates=None):
+        if fieldUpdates is None:
+            cleaned_doc = {k: v for k, v in doc.items() if not self._is_null_value(v)}
+        else:
+            # id must be added without a modifier
+            # if using field updates, all other fields should have a modifier
+            cleaned_doc = {
+                k: {fieldUpdates[k]: v} if k in fieldUpdates else v
+                for k, v in doc.items()
+            }
+
         return cleaned_doc
+            
 
     def _build_xml_doc(self, doc, boost=None, fieldUpdates=None):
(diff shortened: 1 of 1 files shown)


acdha (approved):

acdha:
This looks like a good idea — it's been a while but I believe we also will see a non-trivial performance win moving away from XML. The main thing I was thinking we should have are some tests for that conditional logic.

mitchelljkotler:
Great!  Are you interested in completely removing XML?  Is index time boosting something important to keep available?

Otherwise, I think the best way to test the conditional logic would be to refactor it out into another method - right now its embedded within the add method and would be hard to test.  I can work on that.

PR #391, opened 2023-04-25
Simplify using flake8-simplify

cclauss:
% `ruff --select=SIM108,SIM118 --fix .`
```
Found 7 errors (7 fixed, 0 remaining).
```
% `ruff rule SIM108`
# if-else-block-instead-of-if-exp (SIM108)

Derived from the **flake8-simplify** linter.

Autofix is sometimes available.

Message formats:
* Use ternary operator `{contents}` instead of `if`-`else`-block

% `ruff rule SIM118`
# in-dict-keys (SIM118)

Derived from the **flake8-simplify** linter.

Autofix is always available.

Message formats:
* Use `{key} in {dict}` instead of `{key} in {dict}.keys()`

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -708,10 +708,7 @@ class Solr(object):
             else:
                 value = "%sT00:00:00Z" % value.isoformat()
         elif isinstance(value, bool):
-            if value:
-                value = "true"
-            else:
-                value = "false"
+            value = "true" if value else "false"
         else:
             if IS_PY3:
                 # Python 3.X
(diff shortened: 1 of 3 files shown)


pysolr.py
```diff
@@ -708,10 +708,7 @@ def _from_python(self, value):
             else:
                 value = "%sT00:00:00Z" % value.isoformat()
         elif isinstance(value, bool):
-            if value:
-                value = "true"
-            else:
-                value = "false"
+            value = "true" if value else "false"
```
acdha, line 711: I would disable this. It’s harder to read, generates more diff noise if you change either branch, and vertical height isn’t a strictly limited resource.

cclauss:
Closing in favor of #392

PR #394, opened 2023-04-26
Replace flake8 and isort with ruff

cclauss:
[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

The ruff Action uses minimal steps to run in ~5 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image

Running the complete test matrix on GitHub Actions takes a lot of time and CPU.  Given that the `ruff` linting job is so fast (~5 seconds) let's ensure that linting passes ___before___ the complete test matrix is run.  This will cost a few seconds on each run but could save a lot of time and contention for PRs that should be cleaned up before they are fully tested.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds
image

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,11 +1,20 @@
-name: tox
+name: ruff_and_tox
 on:
   push:
     branches: [master]
   pull_request:
     branches: [master]
 jobs:
+  ruff:  # https://beta.ruff.rs
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: pip install --user "codespell[toml]" ruff
+    - run: codespell
+    - run: ruff --format=github .
+
   tox:
+    needs: ruff
     strategy:
       fail-fast: false
       max-parallel: 6
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,4 @@
 repos:
-    - repo: https://github.com/PyCQA/isort
-      rev: 5.12.0
-      hooks:
-          - id: isort
-    - repo: https://github.com/psf/black
-      rev: 23.3.0
-      hooks:
-          - id: black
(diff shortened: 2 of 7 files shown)


.github/workflows/tox.yml
```diff
@@ -1,11 +1,20 @@
-name: tox
+name: ruff_and_tox
```
acdha, line 1: I’m wondering whether we should split this into separate code quality and tests files since there really isn’t a need to run Ruff as a blocker for tox, but it also isn’t that important.

acdha (approved):

acdha:
Looks like we need codespell to ignore the `.git` directory

acdha:
Thanks for doing this! I think we get the last couple spellcheck items in & then do a second PR where we enable the pyupgrade fixes and remove the remaining Python 2 compatibility code.

.github/workflows/tox.yml
cclauss: If codespell + ruff took more than 10 seconds then I would not run them as a precursor but given that they are so fast that it is a painless precaution that often saves time and energy.

acdha: Yeah, I think the job setup & teardown takes longer than the actual ruff execution.

PR #396, opened 2023-04-26
Replace ''.join after char filtering with regex filter

acdha:
Rebased & consolidated version of #257

Closes #257

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -83,6 +83,10 @@ DATETIME_REGEX = re.compile(
 # dict key used to add nested documents to a document
 NESTED_DOC_KEY = "_childDocuments_"
 
+VALID_XML_CHARS_REGEX = re.compile(
+    "[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD\U00010000-\U0010FFFF]+"
+)
+
 
 class NullHandler(logging.Handler):
     def emit(self, record):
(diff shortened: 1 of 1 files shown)


cclauss:
```python
"""
s = "".join(chr(i) for i in range(0x110000))
len(s)       = 1,114,112
len(old_str) = 1,112,033
len(new_str) = 1,112,033
old: 5.4839835410239175
new: 0.31367145804688334
"""
import re
from timeit import timeit


def is_valid_xml_char_ordinal(i):
    return (
        0x20 <= i <= 0xD7FF
        or i in (0x9, 0xA, 0xD)
        or 0xE000 <= i <= 0xFFFD
        or 0x10000 <= i <= 0x10FFFF
    )


def old_way(s):
    return "".join(c for c in s if is_valid_xml_char_ordinal(ord(c)))


VALID_XML_CHARS_REGEX = re.compile(
    "[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD\U00010000-\U0010FFFF]+"
)


def new_way(s):
    return VALID_XML_CHARS_REGEX.sub("", s)


s = "".join(chr(i) for i in range(0x110000))
old_str = old_way(s)
new_str = new_way(s)

print(f"{len(s)       = :,}")
print(f"{len(old_str) = :,}")
print(f"{len(new_str) = :,}")

assert len(old_str) == len(new_str)
assert old_str == new_str

print("old:", timeit("old_way(s)", setup="from __main__ import old_way, s", number=20))
print("new:", timeit("new_way(s)", setup="from __main__ import new_way, s", number=20))
```

cclauss (approved):

PR #435, opened 2023-10-14
Make bool values lowercase in solr query url - fixes #401

### LINKED ISSUE #401: Unable to retrieve results using pysolr.search() with grouping
# I have
* [* ] Tested with the latest release
* [ ] Tested with the current master branch
* [* ] Searched for similar existing issues

## Expected behaviour
I expect to retrieve search results grouped by the specified field (**sno**) based on the provided query. The results should be similar to what I see in the Solr UI.

## Actual behaviour
No search results are returned. The results variable is empty.

## Steps to reproduce the behaviour
1. Use pysolr library to connect to Solr instance.
2. Execute the following search query with grouping in python:
    results = solr.search(query, **{"group": "true", "group_field": "sno", "group_offset": 0, "group_limit": 1, "group_sort": "date desc", "group_ngroups": "true", "rows": 1000})

Alternatively I tried below query that gives me _(HTTP 400): [Reason: invalid boolean value: True]_- 
results = solr.search(query, **{_"group":True_, "group_field": "sno", "group_offset": 0, "group_limit": 1, "group_sort": "date desc", _"group_ngroups": True,_ "rows": 1000})

## Configuration

* Operating system version: Windows 10
* Search engine version:
* Python version: 3.8
* pysolr version: 3.9.0

ch2ohch2oh:
Minor changes as titled. Also added a small util function to look up nested json objects.

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -191,6 +191,9 @@ def safe_urlencode(params, doseq=0):
     which can't fail down to ascii.
     """
     if IS_PY3:
+        for key, val in params.items():
+            if isinstance(val, bool):
+                params[key] = str(val).lower()
         return urlencode(params, doseq)
 
     if hasattr(params, "items"):
@@ -298,6 +301,16 @@ class Results(object):
             for d in result.docs:
                 yield d
             result = result._next_page_query and result._next_page_query()
+
+
+def get_nested(obj, keys, default=None):
+    """Nested key lookup for a dict-like object."""
+    try:
+        for k in keys:
+            obj = obj[k]
+        return obj
+    except KeyError:
+        return default
 
 
 class Solr(object):
@@ -528,7 +541,7 @@ class Solr(object):
         path_handler = handler
         if self.use_qt_param:
             path_handler = "select"
-            query_vars.append("qt=%s" % safe_urlencode(handler, True))
+            query_vars.append("qt=%s" % handler)
 
         path = "%s/" % path_handler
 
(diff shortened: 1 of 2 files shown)


pysolr.py
```diff
@@ -300,6 +303,16 @@ def __iter__(self):
             result = result._next_page_query and result._next_page_query()
 
 
+def get_nested(obj, keys, default=None):
```
acdha, line 306: It looks like this is only used by a debug logging call, and it could be replaced by a standard dictionary get-with-default call there.

pysolr.py
```diff
@@ -528,7 +541,7 @@ def _update(
         path_handler = handler
         if self.use_qt_param:
             path_handler = "select"
-            query_vars.append("qt=%s" % safe_urlencode(handler, True))
```
acdha, line 531: This appears to be unrelated to the proposed change.

pysolr.py
```diff
@@ -830,11 +843,17 @@ def search(self, q, search_handler=None, **kwargs):
         response = self._select(params, handler=search_handler)
         decoded = self.decoder.decode(response)
 
-        self.log.debug(
-            "Found '%s' search results.",
-            # cover both cases: there is no response key or value is None
-            (decoded.get("response", {}) or {}).get("numFound", 0),
-        )
+        if decoded.get("grouped"):
```
acdha, line 846: I would defer support for grouped responses to a larger task which would provide more help for them - just changing a log message doesn't seem to add much value.

tests/test_client.py
```diff
@@ -72,6 +73,9 @@ def test_safe_urlencode(self):
             "test=Hello \u2603!&test=Helllo world!",
         )
 
+        # Boolean options for Solr should be in lowercase.
+        self.assertTrue("True" not in safe_urlencode({"group": True}))
```
acdha, line 77: This would pass a bug which encoded it as `group=None` or `group="false"`.

```suggestion
        self.assertEqual("group=true", safe_urlencode({"group": True}))
```

pysolr.py
```diff
@@ -191,6 +191,9 @@ def safe_urlencode(params, doseq=0):
     which can't fail down to ascii.
     """
     if IS_PY3:
+        for key, val in params.items():
```
acdha, line 194: In general, pysolr has avoided doing type coercion to avoid needing to know what type(s) Solr supports for a given field (e.g. a parameter like `group` is documented as accepting `true` but will also accept `on` or `yes` but not `1` or `True`) but I think this is relatively safe because Solr doesn't accept that value natively so the only case where this could cause a problem is if someone was passing a Solr string which they for some reason wanted to have processed literally — for example, if I had a Python app which used a Solr StringField for something and expected the literal value `True` or `False` because that field type isn't case-insensitive.

cclauss:
git conflicts — please rebase.

PR #490, opened 2024-11-02
Fix extraction in Solr >= 8.

andresmrm:
Related to https://github.com/django-haystack/django-haystack/pull/2000

Didn't run the test suite...

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -1284,10 +1284,21 @@ class Solr(object):
             self.log.exception("Failed to load JSON response")
             raise
 
-        data["contents"] = data.pop(filename, None)
+        if "file" in data:
+            # For Solr >= 8.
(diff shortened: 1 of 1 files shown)


DhavalGojiya:
> I think this comes down to what you think of as a breaking change: if this required you to update your Python code, it'd definitely become 4.0.0 but I'm not sure dropping support for version of Solr which hasn't been supported in years is the same.

Yeah, we are going to change/update/add only the test case and assertions with new Solr versions after dropping old Solr version support. The actual Pysolr package source code remains unchanged, so a new minor bump to `pysolr==3.12.0` will still work without any breaking changes.

DhavalGojiya:
> @andresmrm Please resolve git conflicts.

The following changes in this PR are already covered by my Solr upgrade PR.

Issue #515, opened 2025-11-06
Ruff version mismatch and outdated configuration


DhavalGojiya:
The current `ruff` version in `Pipfile.lock` is `0.0.269`, which is different from the version used in the `.pre-commit-config.yaml` file — the latest one being `0.14.3`.

Because of this older version, Ruff fails to recognize the new `[tool.ruff.lint]` configuration syntax.  
Running `ruff check` after setting up the project throws a TOML parse error related to the `lint` field.

### 1. With Old Version `0.0.269`

```shell
(pysolr) dhaval@lg-ubuntu:~/Open-Source/pysolr$ ruff --version
ruff 0.0.269

(pysolr) dhaval@lg-ubuntu:~/Open-Source/pysolr$ ruff check .
error: TOML parse error at line 4, column 1
  |
4 | lint.select = [
  | ^^^^
unknown field `lint`, expected one of ...
```

### 2. With Latest Version `0.14.3`

```shell
(pysolr) dhaval@lg-ubuntu:~/Open-Source/pysolr$ uvx ruff@0.14.3 check .
All checks passed!
```

---

### ✅ Expected Behaviour

We need to update the `ruff` version to the latest in the `Pipfile.lock` as well.

PR #516, opened 2025-11-06
fix: Remove ruff (handled by pre-commit.ci, not a runtime dependency)

DhavalGojiya:
Fixes [#515](https://github.com/django-haystack/pysolr/issues/515)

- Remove ruff dependency from the project (handled by pre-commit.ci, not a runtime dependency)
- Remove ruff job from the github action workflow

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,4 +1,4 @@
-name: CI
+name: ci
 on:
   push:
     branches: [master]


DhavalGojiya:
> Nice, but let's do this a bit differently. `pre-commit.ci` is doing its thing, so we don't need to repeat it in GitHub Actions.
> 
> Please remove from `pyproject.toml` all tools that are run in pre-commit. (They are not runtime dependencies or even test-time dependencies.)
> 
> Then let's remove the `ruff` job at:
> 
> https://github.com/django-haystack/pysolr/blob/39351f36634c906c123a1d4d0676a00adf9793b7/.github/workflows/tox.yml#L7-L17
> 
> Make sure that you drop the `needs: ruff` on line 17.
> 
> Also, please change line 1 to name this either `ci` or `tox`.

Suggestions updated.

cclauss (approved):
Perfect.  Thanks for doing this.

PR #518, opened 2025-11-10
EOL: Drop support for Python 3.9 and below

DhavalGojiya:
### Summary

As of 2025-10, Python 3.9 is end of life and does not get upstream support from python.org anymore.

This PR drops support for **Python 3.9 and below** to align with the current Python ecosystem and simplify code maintenance.

Python 3.9 reached end of life: [https://devguide.python.org/versions/](https://devguide.python.org/versions/)

---

### Changes

- [x] Added `python_requires=">=3.10"` in `setup.py`
- [x] Removed references and compatibility for older Python versions
- [x] Update project manager `Pipenv` with Python 3.10+
- [x] Update GitHub Actions CI/CD workflow to drop Python 3.9
- [x] Update README

---

### Notes

Users on Python versions earlier than 3.10 will now receive a version restriction error when trying to install Pysolr version 3.11 or higher via pip.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ setup(
     ],
     url="https://github.com/django-haystack/pysolr/",
     license="BSD",
-    install_requires=["requests>=2.9.1", "setuptools"],
+    install_requires=["requests>=2.32.5", "setuptools"],
     python_requires=">=3.10",
     extras_require={"solrcloud": ["kazoo>=2.5.0"]},
     setup_requires=["setuptools_scm"],
--- a/README.rst
+++ b/README.rst
@@ -25,8 +25,8 @@ Features
 Requirements
 ============
 
-* A supported version of Python 3
-* Requests 2.9.1+
+* Python 3.10+
+* Requests 2.32.5+
 * **Optional** - ``simplejson``
 * **Optional** - ``kazoo`` for SolrCloud mode
 
@@ -51,9 +51,6 @@ Usage
 Basic usage looks like:
 
 .. code-block:: python
-
-    # If on Python 2.X
-    from __future__ import print_function
 
     import pysolr
 
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,5 +1,3 @@
-# encoding: utf-8
-
 from .test_admin import *  # NOQA
 from .test_client import *  # NOQA
 from .test_cloud import *  # NOQA


DhavalGojiya:
> Let's plan to tag this as **3.11** or perhaps **4.0**, so it's easy to backport something like a security update.

I strongly recommend that we **bump the major version**, e.g. `pysolr 4.0.0`

Here are my points to support this statement:

* A **major version bump** directly indicates that we introduced a significant change.
* Suppose tomorrow Apache Solr introduces a new endpoint to work with core or schema - we will likely provide support for it in `pysolr`, right?
  In most cases, such features can also be implemented using the `requests` module or other built-in tools, meaning backporting them to Python 3.9 - supported versions (like `pysolr 3.10.x` or `3.11.x`) is still possible.
* By bumping to **`pysolr 4.0.0`**, we create a clear separation from the `3.x` series (`3.10.x`, `3.11.x`, `3.12.x`, etc.) that still supports Python 3.9. This allows us not only to backport security fixes but also to backport **new features** (via minor bumps) to those versions - while keeping `4.x` focused on modern Python support.
* I am not saying that we must also backport possible new feature changes to the old 3.x.x series - as @cclauss clarified a lot of things - but we'll still have the option to do so if needed. For users running Python versions below 3.10, at least we'll have the flexibility to consider it.

acdha:
> That is _**paid**_ support. So if users are paying Red Hat or Amazon, then those organizations can provide the required support.

It's not necessarily paid support – only one of the examples I mentioned requires payment – but to your larger point, yes, I think we should drop support following the upstream CPython support lifecycle. I just want to do it in a way which leaves an easy pinning option for anyone depending on pysolr in a project which still supports users on those long-term stable Linux distributions.  

> Here are my points to support this statement: …

Was some of this written with the help of an LLM? Some of the hypotheticals or the way `requests` is referenced seem generic in that style and I'm not sure they're germane to a small project like Pysolr. I really don't anticipate the need to back port new features, for example, since the only reason we'd touch it would be something like an exploitable security problem.

My vote would be using a minor release (3.11). That's not a strong argument against a major release because the cost of either is low, but this isn't really changing the actual public API and as long as the package is correctly tagged people can't easily install new releases on Python 3.9. Historically we haven't had problems caused by dropping support for old Python versions in minor releases.

Pipfile
```diff
@@ -9,6 +9,7 @@ coverage = "*"
 [packages]
 requests = "*"
 kazoo = "*"
+six = "*"
```
DhavalGojiya: The `six` package is added here explicitly because the test cases depend on it.  

Earlier, it was automatically installed by Kazoo, but since we updated Kazoo, they have removed `six` as a dependency in the newer version.  
Therefore, it has been added here explicitly.

DhavalGojiya:
Strange!  
Why is the GitHub Actions workflow suddenly taking more time?

acdha:
> > Was some of this written with the help of an LLM?
> 
> No. I asked this in the official Django Discord server, and that's where I got my some supporting points.

Ah, cool, I just wasn't sure where the idea of things like back porting features came from since Pysolr is unlike a lot of other projects in that it's really just a wrapper for upstream functionality so there's much more limited scope for that kind of thing. Django Haystack might have a little more of that since there are people who might be stuck with other dependencies blocking, say, a major Django or Python upgrade but this project is mostly a shim.

Pipfile
DhavalGojiya: Done :)

setup.py
```diff
@@ -23,11 +23,8 @@
     ],
     url="https://github.com/django-haystack/pysolr/",
     license="BSD",
-    install_requires=[
-        "requests>=2.9.1",
-        "setuptools",
-        "importlib_metadata; python_version<'3.8'",
-    ],
+    install_requires=["requests>=2.9.1", "setuptools"],
```
acdha: requests 2.9.1 is pretty old (2015). I was wondering whether we might want to bump this to `requests>=2.32` since that's technically needed to get support for Python 3.14 (2.32.5) and if we don't pin the minor version that's a release which came out over a year ago (2.32.0 was May 2024).

acdha (approved):
I think we could go with this immediately. If we want to adjust the requests minimum version that can happen later.

DhavalGojiya:
> This is looking like a very nice cleanup PR. I'm enjoying removal of all of the Python 2 checks, thank you!

Welcome 🤓

DhavalGojiya:
> I think we could go with this immediately. If we want to adjust the requests minimum version that can happen later.

Ok, but we also need to update the README file. I can see that some code blocks in the README mention `Python 2.x`.

PR #522, opened 2025-11-17
PEP 621: Migrate from setup.py to pyproject.toml

cclauss:
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
* https://peps.python.org/pep-0621

Migrate `setup.py` to `setup.cfg` using [setuptools-py2cfg](https://pypi.org/project/setuptools-py2cfg).  Then migrate `setup.cfg` to `pyproject.toml` using [ini2toml](https://pypi.org/project/ini2toml) to do the file conversion, and run [pyproject-fmt](https://pypi.org/project/pyproject-fmt) and then [validate-pyproject](https://github.com/abravalheri/validate-pyproject) to validate the results.

```bash
uvx setuptools-py2cfg > setup.cfg
uvx 'ini2toml[full]' setup.cfg -o new_pyproject.toml
code *.toml  # Prepend the contents of new_pyproject.toml to the beginning of pyproject.toml
rm new_pyproject.toml
git rm setup.py
rm setup.cfg
uvx pyproject-fmt pyproject.toml
uvx 'validate-pyproject[all]' pyproject.toml
git grep setup
code sonar-project.properties README.rst
git commit -am"PEP 621: Migrate from setup.py to pyproject.toml"
git push
```
---
Related to:
* #521 because
> If you're already using a `pyproject.toml` file, we recommend `project.requires-python` instead, as it's based on Python packaging standards, and will be respected by other tools. -- https://docs.astral.sh/ruff/settings/#target-version


DhavalGojiya:
@cclauss
1 - Why removed the whole build backend table? (since it Defaults fallback for most of front end build command?) 
2 - It's still unclear that we should keep "setuptools" as base dependency of pysolr? As build backend `requires` are enough to built source distribution.

DhavalGojiya:
What happened? @cclauss

PR #561, opened 2026-04-15
chore: Clean up unnecessary files from the repository that are no longer in use

DhavalGojiya:
## Description

Clean up unnecessary files from the repository that are no longer in use.

### Removed Files:
- .gitchangelog.rc
- sonar-project.properties


DhavalGojiya:
@acdha Your review plz.

acdha:
We still use .editorconfig – many editors will use it to auto-configure their settings to match: https://editorconfig.org/#pre-installed

Otherwise this looks reasonable.

DhavalGojiya:
> We still use .editorconfig – many editors will use it to auto-configure their settings to match: https://editorconfig.org/#pre-installed
> 
> Otherwise this looks reasonable.

Done :)

cclauss (approved):

PR #563, opened 2026-04-20
chore: Remove legacy code from pysolr module and move metadata to pyproject.toml

DhavalGojiya:
## Description

- Clean up versioning code in the pysolr module, as dynamic versioning is no longer used.
- Remove author metadata from the pysolr module and move it to pyproject.toml. See: [python-packaging-doc](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#authors-maintainers)
- Move the `coverage` dependency to the test dependency group, as it belongs to testing.

### Extra
* _Package Screenshot_
Pasted image


cclauss (approved):

DhavalGojiya:
Can we add @cclauss and @acdha to the authors or maintainers list? You both have also put in a lot of hard work here.

cclauss:
https://github.com/django-haystack/pysolr/blob/de19c6a6cbcb677dfa75f2ef83c0362e728e9cdf/AUTHORS#L6

DhavalGojiya:
> https://github.com/django-haystack/pysolr/blob/de19c6a6cbcb677dfa75f2ef83c0362e728e9cdf/AUTHORS#L6

`pyproject.toml` already provides separate sections for authors and maintainers. I think it's better to list these names in the TOML file rather than having a separate AUTHORS file.

pyproject.toml
```diff
@@ -8,7 +8,12 @@ version = "3.11.0"
 description = "Lightweight Python client for Apache Solr"
 readme = "README.md"
 license = { text = "BSD" }
-authors = [ { name = "Daniel Lindsley", email = "daniel@toastdriven.com" } ]
+authors = [
```
acdha, line 11: I don't object to adding more people but I wonder whether this should be a longer list or pruned down to the recently-active maintainers - if we include everyone who's contributed, `git shortlog -n -s` has roughly 70 entries excluding bots. If we only look at the last couple of years, that's much smaller:

```bash
~/Projects/pysolr (master)> git shortlog -n -s --since=2024-04-21
    40  Dhaval Gojiya
    17  Chris Adams
     5  Christian Clauss
     1  David Hotham
     1  Jan Rydzewski
```

DhavalGojiya: Good point.
Currently, it looks like only the three of us (cclauss, you, and me) have been maintaining this project for quite a long time and will likely continue to do so.
We can add (cclauss, you, and me) to the Authors section in `pyproject.toml`, along with the original author, Daniel Lindsley.

We can update the AUTHORS file to include all contributors, or remove it entirely, given that most are not active maintainers anymore, if you prefer.

acdha: I feel like `AUTHORS` files are less relevant in the Git-era where everyone has a full copy of the committers list.

acdha (approved):

pyproject.toml
DhavalGojiya: @acdha Yes definitely.

PR #564, opened 2026-04-22
refactor: Remove legacy safe_urlencode wrapper function from the code

DhavalGojiya:
## Description

1. Remove the legacy `safe_urlencode` wrapper around `urllib.parse.urlencode`.
2. Fix the Solr class docstring URL, which previously did not include the core name, it has now been updated to include it. The README has been updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Basic usage looks like:
 import pysolr
 
 # Create a client instance. The timeout and authentication options are not required.
-# Solr URL format: http://host:port/solr/<CORE_NAME>
+# Solr URL format: http://host:port/solr/<core_name>
 solr = pysolr.Solr("http://localhost:8983/solr/my_core", always_commit=True, [timeout=10], [auth=<type of authentication>])
 
 # Note that auto_commit defaults to False for performance. You can set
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -79,7 +79,7 @@ Basic usage looks like:
 import pysolr
 
 # Create a client instance. The timeout and authentication options are not required.
-# Solr URL format: http://host:port/solr/<CORE_NAME>
+# Solr URL format: http://host:port/solr/<core_name>
```
cclauss, line 82: This should match other occurrences:
https://github.com/search?q=repo%3Adjango-haystack%2Fpysolr+core_name&type=code

Should we standardize: `solr_core_name` or `core_name` or `my_core`?  I like the first, but am flexible.

cclauss (dismissed):

README.md
DhavalGojiya: > Should we standardize: `solr_core_name` or `core_name` or `my_core`? I like the first, but am flexible.

I think this is already fine: `http://host:port/solr/<core_name>`.

URLs are typically written in lowercase for consistency, and `<core_name>` clearly indicates that the user should replace it with the actual core name.

As an example, I used `my_core` as the core name. However, we could avoid using `my_core` and instead use `<core_name>` everywhere to avoid any confusion or guessing.

cclauss (approved):

acdha (approved):

PR #573, opened 2026-06-05
chore(module): restructure the main pysolr module

DhavalGojiya:
## Description

- Restructured the main pysolr module
- Grouped constants, functions, and classes together

### DIFF
--- a/pysolr.py
+++ b/pysolr.py
@@ -29,11 +29,49 @@ from urllib.parse import quote, urlencode
 __all__ = ["Solr"]
 
 
+# ---------------------------
+# Constants
+# ---------------------------
 DATETIME_REGEX = re.compile(
     r"^(?P<year>\d{4})-(?P<month>\d{2})-(?P<day>\d{2})T(?P<hour>\d{2}):(?P<minute>\d{2}):(?P<second>\d{2})(\.\d+)?Z$"  # NOQA: E501
 )
+
 # dict key used to add nested documents to a document
 NESTED_DOC_KEY = "_childDocuments_"
+
+# Using two-tuples to preserve order.
+REPLACEMENTS = (
+    # Nuke nasty control characters.
+    (b"\x00", b""),  # Start of heading
+    (b"\x01", b""),  # Start of heading
+    (b"\x02", b""),  # Start of text
+    (b"\x03", b""),  # End of text
+    (b"\x04", b""),  # End of transmission
+    (b"\x05", b""),  # Enquiry
+    (b"\x06", b""),  # Acknowledge
+    (b"\x07", b""),  # Ring terminal bell
+    (b"\x08", b""),  # Backspace
+    (b"\x0b", b""),  # Vertical tab
+    (b"\x0c", b""),  # Form feed
+    (b"\x0e", b""),  # Shift out
+    (b"\x0f", b""),  # Shift in
+    (b"\x10", b""),  # Data link escape
+    (b"\x11", b""),  # Device control 1
+    (b"\x12", b""),  # Device control 2
+    (b"\x13", b""),  # Device control 3
+    (b"\x14", b""),  # Device control 4
+    (b"\x15", b""),  # Negative acknowledge
+    (b"\x16", b""),  # Synchronous idle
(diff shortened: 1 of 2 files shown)


DhavalGojiya:
Your review plz : @cclauss

cclauss:
Why not use [`str.maketrans()`](https://docs.python.org/3/library/stdtypes.html#str.maketrans) and `str.translate()`?

DhavalGojiya:
> Why not use [`str.maketrans()`](https://docs.python.org/3/library/stdtypes.html#str.maketrans) and `str.translate()`?

Done :)

spent some time learning about these methods.
Thanks @cclauss. Every day, I am learning something new from you. 😅

DhavalGojiya:
Your review plz: @acdha

pysolr.py
```diff
@@ -29,12 +29,24 @@
...
+# ---------------------------
 DATETIME_REGEX = re.compile(
     r"^(?P<year>\d{4})-(?P<month>\d{2})-(?P<day>\d{2})T(?P<hour>\d{2}):(?P<minute>\d{2}):(?P<second>\d{2})(\.\d+)?Z$"  # NOQA: E501
 )
+
 # dict key used to add nested documents to a document
 NESTED_DOC_KEY = "_childDocuments_"
 
+# All C0 controls except tab, newline and carriage return, which are valid whitespace.
+_CONTROL_CHARS = "".join(
```
acdha: I wouldn't use an _ for a module-level constant — if someone tries to redefine it and it breaks, that's on them — but also if we want to avoid leaking a temporary variable, the easiest way to do that would be to put it into the only place it's used:

```python
SANITIZED_CHARS_TABLE = str.maketrans({chr(code): "" for code in range(0x20) if code not in (0x09, 0x0A, 0x0D)})
```

We could also use a regular expression for this but `str.translate` is roughly twice as fast on long strings so I think it makes sense to retain it.

pysolr.py
```diff
@@ -119,10 +134,27 @@ def clean_xml_string(s):
...
+
+# ---------------------------
+# Exceptions
+# ---------------------------
 class SolrError(Exception):
     pass
 
 
+# ---------------------------
+# Classes
```
acdha: I generally don't like group comments like this because it doesn't explain why they're in a group — e.g. the immediately preceding `SolrError` is a class, too — and was thinking it might make sense to have the logical grouping not repeating things which the names & most editors will tell you anyway but more grouped by function: i.e. the core Pysolr classes everyone uses first and then the stuff which only people working directly with SolrCloud have to think about.

DhavalGojiya: Done :)

pysolr.py
DhavalGojiya: Done :)

pysolr.py
```diff
@@ -643,7 +655,11 @@ def _to_python(self, value):
             is_string = True
 
         if is_string:
-            possible_datetime = DATETIME_REGEX.search(value)
+            datetime_regex = re.compile(
```
acdha: This should be back at the module level: that compiles once in the lifetime of the program as opposed to every call.

DhavalGojiya: Reverted :)

pysolr.py
```diff
@@ -116,7 +113,26 @@ def clean_xml_string(s):
...
+
+def sanitize(data):
+    """
+    Removes invalid control characters from ``data``.
+
+    The value is coerced to a Unicode string and all C0 control characters
+    are stripped, except for tab, newline and carriage return, which are
+    valid whitespace.
+    """
+    sanitized_chars_table = str.maketrans(
```
acdha, line 131: This should be up at the module level so it's not run on every call.

DhavalGojiya: >  if we want to avoid leaking a temporary variable, the easiest way to do that would be to put it into the only place it's used:

acdha: Sorry if there was any confusion, my example was at the module level combining the `_CONTROL_CHARS` definition and `maketrans` result into a single module-level declaration.

DhavalGojiya: Done :)

acdha (approved):

PR #591, opened 2026-09-16
refactor(test): assert commitWithin request param instead of commit timing

DhavalGojiya:
## Description

See -> https://github.com/django-haystack/pysolr/actions?query=is%3Afailure+workflow%3Aci

Instead of a time-based flaky check, we now check and make sure the HTTP request is created correctly and assert that it contains the expected values.

When the document is actually indexed and becomes searchable, that is Solr's responsibility.

### Test
image


DhavalGojiya:
@cclauss 
If everything looks good to you, I can make the final commit, which will remove the 200 runs that we added to verify our flaky test fix.

cclauss (dismissed):

cclauss:
Approved when you are ready to merge.

If you are not ready to have a pull request merged, please open it as a DRAFT.
* https://github.blog/news-insights/product-news/introducing-draft-pull-requests

tests/base.py
```diff
@@ -648,20 +648,29 @@ def test_add_with_boost(self):
...
             ],
             commitWithin=commit_within_ms,
         )
+
         # we should not see the doc immediately
         assert len(self.solr.search("commitWithin")) == 0
+
         # but we should see it after commitWithin period (+ small grace period)
-        time.sleep((commit_within_ms / 1000.0) + 0.01)
+        time.sleep((commit_within_ms / 1000.0) + 0.5)
```
acdha: Tests like this often end up being fragile and I was wondering whether we should have what is fundamentally a test of Solr's behaviour rather than simply using mock to confirm that the `commitWithin` value passed to `solr.add` is included in the HTTP request.

DhavalGojiya: @acdha 
Yeah, completely agree with you. Timing-related tests depend on a lot of other factors, so they can be pretty fragile.

It's not really pysolr's responsibility how long Solr takes to make a document searchable. We should just test that the HTTP request is built correctly and that the `commitWithin` value is passed correctly.

Working on it.

DhavalGojiya: Done :)

acdha (approved):

PR #593, opened 2026-09-17
chore(packaging): add free-threading trove classifier

DhavalGojiya:
## Description

`pysolr` supports the free-threaded Python 3.14t build, and we are also testing it in CI. So, add the free-threading classifier to the package metadata.

See: https://py-free-threading.github.io/porting/#porting-python-packages-to-support-free-threading


cclauss:
I think we first need to test with [`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel).
* https://py-free-threading.github.io/testing

DhavalGojiya:
> I think we first need to test with [`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel).

The only problem with running `pysolr` test cases in parallel is that we have a single Apache Solr instance (Java) running at the system level.

If we run the `pysolr` test cases in parallel, I am not sure about the Python-level behavior, but both test cases may perform index/remove/update operations on the same Solr instance. So, failures could happen at the Solr system level rather than where we expect them. We need to check how to handle these cases.

cclauss:
In a separate PR, please test on Python 3.15t because we are just two weeks from the production release.

cclauss:
Please rebase.

PR #594, opened 2026-09-17
chore(pre-commit): add actionlint hook for GitHub Actions workflows

DhavalGojiya:
## Description

Add `actionlint` pre-commit git hook.

See: https://github.com/rhysd/actionlint/blob/v1.7.12/docs/checks.md


cclauss (approved):
It does not seem to be too chatty.

PR #598, opened 2026-09-17
Replace requests with httpx2

### LINKED ISSUE #597: Upgrade from requests to httpx2
% `git grep requests ` # --> https://github.com/search?q=repo%3Adjango-haystack%2Fpysolr+requests

If we are going to claim we are free-threaded, let's upgrade to https://pydantic.dev/docs/httpx2

@DhavalGojiya

priya-sundaram-dev:
Closes #597.

Swaps `requests` for [`httpx2`](https://github.com/pydantic/httpx2) (the Pydantic-maintained fork of httpx) so the "we're free-threaded" story rests on a currently-maintained HTTP client.

### What changed
- **Dependency:** `requests>=2.32.5` → `httpx2>=2.12` in `pyproject.toml`.
- **Client:** `requests.Session()` → `httpx.Client(verify=..., follow_redirects=True)`. `verify` is set at construction (httpx doesn't honour it as a post-hoc attribute) and `follow_redirects=True` preserves requests' default redirect behaviour.
- **Body handling:** requests accepted both raw bytes and form dicts under `data=`; httpx splits these. Raw XML/JSON bodies now go through `content=`, while the multipart `extract()` path (dict fields + `files=`) uses `data=`. `content`/`files` are only passed to verbs that take a body, since httpx's per-method helpers reject them on `GET`.
- **Exceptions:** `Timeout`→`httpx.TimeoutException`, `ConnectionError`→`httpx.ConnectError`, `HTTPError`→`httpx.HTTPStatusError`, `RequestException`→`httpx.RequestError`, and JSON-decode failures caught via `ValueError` (httpx's `.json()` raises `json.JSONDecodeError`).
- **Docs/tests:** README custom-auth example now uses [`httpx-gssapi`](https://github.com/pythongssapi/httpx-gssapi) (the httpx counterpart to `requests-kerberos`); `test_client` asserts the chained cause is `httpx.ConnectError`; CHANGELOG updated.

### Behaviour changes worth a maintainer's eye
- A user-supplied `session=` must now be an `httpx.Client`, not a `requests.Session`.
- Kerberos/SPNEGO users move from `requests-kerberos` to `httpx-gssapi`.
- A custom `auth=` object must be httpx-compatible (a `(user, pass)` tuple still works).

### Testing
`ruff check` / `ruff format --check` are clean. I exercised each `_send_request` path (GET with no body, POST with a raw body, the `extract` multipart path, the admin `raise_for_status`/JSON-decode branches, and 500 → `SolrError`) against an `httpx.MockTransport`, plus a real connection-refused check to confirm `SolrError.__cause__` is an `httpx.ConnectError`. The Solr integration suite needs a live Solr, so I couldn't run it here — happy to iterate if CI surfaces anything.

I kept the diff focused on the request/response layer. Let me know if you'd prefer `httpx2>=2.12` pinned differently.


tests/test_client.py
```diff
@@ -1,8 +1,8 @@
 from unittest.mock import Mock
 from urllib.parse import unquote_plus, urlencode
 
+import httpx2 as httpx
```
cclauss: Awesome (as always)!!

My strong preference would be to `import httpx2` and use `httpx2` everywhere. Initially, it will be highly compatible with the original, but sooner or later Pydantic will evolve in ways that may break backward compatibility. I just don’t want visitors to misread the code and assume it is using the original `httpx`.

README.md
```diff
@@ -193,24 +193,26 @@ handler in `search` explicitly overrides the `search_handler` setting
 
 ### Custom Authentication
 
+pysolr now uses [httpx2](https://github.com/pydantic/httpx2) as its HTTP
+client, so the `auth` argument accepts any httpx-compatible authentication
+object (or a plain `(username, password)` tuple). For Kerberos/SPNEGO, use
+[httpx-gssapi](https://github.com/pythongssapi/httpx-gssapi) (the httpx
+counterpart to `requests-kerberos`):
```
cclauss, line 200: * pythongssapi/httpx-gssapi#57

cclauss (changes requested):

cclauss:
@DhavalGojiya, would you be willing to create a separate pull request to add `httpx2` to our dependencies but LEAVE REQUESTS IN PLACE?

Once that is merged, Priya can rebase this, and more tests should pass.  Thanks.

priya-sundaram-dev:
Thanks for the review, @cclauss! Fully agree on the readability point — I've pushed a commit that drops the alias: it now `import httpx2` and references `httpx2.*` everywhere (client, exception handlers, tests, and the explanatory comments), so nobody misreads it as the original `httpx`.

Sounds good on the rollout plan: once @DhavalGojiya's PR adds `httpx2` to the dependencies (leaving `requests` in place) and merges, I'll rebase this branch on top so the test suite can resolve `httpx2` and more tests pass. I noted the httpx-gssapi#57 reference on the README example — happy to adjust that snippet too once that lands. Just ping me when you'd like the rebase.

DhavalGojiya:
> @DhavalGojiya, would you be willing to create a separate pull request to add `httpx2` to our dependencies but LEAVE REQUESTS IN PLACE?
> 
> Once that is merged, Priya can rebase this, and more tests should pass. Thanks.

Will do.

cclauss:
* #599 merged.  Please rebase.

priya-sundaram-dev:
Rebased onto `master` now that #599 has merged. 🎉

Resolved the `pyproject.toml` conflict by collapsing the dependencies to `httpx2>=2.13` only — since this PR removes every `requests` code path, `requests` is no longer needed at runtime (I picked up the newer `>=2.13` floor from #599 rather than my original `>=2.12`). All the `httpx2` usage, exception handlers, and tests carry over unchanged from the drop-alias commit.

Ready for another look / CI run whenever suits. Happy to also update the README `httpx-gssapi` snippet in this PR or a follow-up, per the httpx-gssapi#57 note.

DhavalGojiya:
@priya-sundaram-dev

If you are touching the `pyproject.toml` file, you also need to run "uv sync".

Any changes to `pyproject.toml` also need to be reflected in the automatically generated `uv.lock` file.

Without this, our pre-commit CI bot will not pass the `uv lock` hook, and none of the CI jobs will run because we also have a `uv.lock` check in the CI pipelines.

pyproject.toml
```diff
@@ -29,10 +29,7 @@ classifiers = [
   "Programming Language :: Python :: 3.15",
   "Topic :: Internet :: WWW/HTTP :: Indexing/Search",
 ]
-dependencies = [
-  "httpx2>=2.13",
-  "requests>=2.32.5",
-]
+dependencies = [ "httpx2>=2.13" ]
```
cclauss:
```suggestion
dependencies = [
  "httpx2>=2.13",
  "requests>=2.32.5",
]
```

cclauss:
I reverted the changes to `pyproject.toml` dependencies.

priya-sundaram-dev:
Pushed the fix for the failing `test_extract` jobs (2ece698).

**Root cause:** httpx2's stricter multipart encoding rejects `str` payloads — a text-mode file or `io.StringIO` now raises `TypeError: Multipart file uploads require 'io.BytesIO', not 'io.StringIO'.`. `requests` accepted either, so this only surfaced after the migration.

**Fix:** `extract()` now reads the file object and encodes text to bytes before building the multipart field, keeping the public API backwards compatible with text file objects (the test suite passes a `StringIO`). Binary file objects pass through unchanged.

I left `pyproject.toml` as you reverted it (httpx2 + requests), so `uv.lock` is untouched and matches master — no `uv sync` needed. CI should now go green.

cclauss (dismissed):
LGTM

tests/test_client.py
acdha: Seconding this, especially because someone might be doing something like typing `from pysolr import http…` in their IDE, have it autocomplete to `httpx` and go looking at the wrong docs.

priya-sundaram-dev:
Good call, @acdha / @cclauss — fully agree on avoiding the `httpx`↔`httpx2` confusion. The code already does `import httpx2` (no alias) and refers to `httpx2.` everywhere in `pysolr.py` and the tests. I just pushed one more tidy-up (2 lines) fixing a stray "httpx client" mention in a test teardown docstring so the exact `httpx2` name is used consistently across the whole diff. CI is green.

pysolr.py
```diff
@@ -1169,11 +1180,18 @@ def extract(self, file_obj, extractOnly=True, handler="update/extract", **kwargs
         }
         params.update(kwargs)
         filename = quote(file_obj.name.encode("utf-8"))
+        # httpx2 multipart uploads require binary content: a text-mode file or
```
acdha: Since this is going to be pysolr 4 anyway, I'm wondering whether we want this or should punt more responsibility to the caller and say they should provide a binary file. In particular, that should be more memory-efficient for large files since this code is going to store two full copies in memory but a binary file can be uploaded in chunks without re-encoding. That should also avoid subtle errors if given a non-UTF-8 file but I sincerely hope that is no longer common.

One way to do that with less impact would be to pass the object through unless it's an instance of io.TextIOBase. I don't think it's worth making a generator to allow streaming but we could document that.

pysolr.py
```diff
@@ -1321,13 +1338,15 @@ def _send_request(self, url, params=None, headers=None):
                 f"Solr returned HTTP error {error_code}. Response body: {error_msg}"
             ) from e
 
-        except requests.exceptions.JSONDecodeError as e:
+        except ValueError as e:
+            # httpx2's ``Response.json()`` raises ``json.JSONDecodeError`` (a
```
acdha: It is tempting to only catch the `JSONDecodeError` to avoid confusing error messages if at any point the code in the `try` block changes in a way which can generate a `ValueError` for a different reason.

CHANGELOG.rst
```diff
@@ -7,6 +7,17 @@ Changelog
...
+- Replace ``requests`` with ``httpx2`` as the HTTP client (closes #597).
+  [Priya Sundaram]
+
+  pysolr now depends on `httpx2 <https://github.com/pydantic/httpx2>`_, the
+  Pydantic-maintained fork of httpx, instead of ``requests``. The ``auth``
+  argument now accepts any httpx-compatible auth object (or an
+  ``(username, password)`` tuple); Kerberos/SPNEGO users should switch from
+  ``requests-kerberos`` to `httpx-gssapi
+  <https://github.com/pythongssapi/httpx-gssapi>`_. A custom ``session=``
+  must now be an ``httpx.Client`` rather than a ``requests.Session``.
```
acdha:
```suggestion
  must now be an ``httpx2.Client`` rather than a ``requests.Session``.
```

tests/test_admin.py
```diff
@@ -17,7 +17,7 @@ def setup_class(cls):
 
     @classmethod
     def teardown_class(cls):
-        """Close the shared requests session (if created) after all tests."""
+        """Close the shared httpx client (if created) after all tests."""
```
acdha:
```suggestion
        """Close the shared httpx2 client (if created) after all tests."""
```

pysolr.py
DhavalGojiya: +1 for this.

We should not use a broader exception catch when we can catch the specific `JSONDecodeError` exception instead. @priya-sundaram-dev

priya-sundaram-dev:
Thanks both — addressed all of this in f218fad:

- **Binary streaming (`extract`, @acdha):** Good call. Now passes the file object straight through to httpx2 unless it's an `io.TextIOBase`, so binary files stream in chunks with no second in-memory copy (and no UTF-8 assumption). Only text streams get read+encoded, which preserves the existing `StringIO`/text-mode behavior the tests exercise.
- **Narrow the JSON except (@acdha, @DhavalGojiya):** Changed the admin-request handler to catch `json.JSONDecodeError` specifically instead of the broad `ValueError`, so an unrelated `ValueError` from a future change in the `try` block won't be mislabeled as a decode failure.
- **CHANGELOG:** applied the `httpx2.Client` suggestion. (The `test_admin.py` teardown docstring already reads "httpx2 client".)

CI should stay green.

pysolr.py
```diff
@@ -1321,13 +1341,17 @@ def _send_request(self, url, params=None, headers=None):
                 f"Solr returned HTTP error {error_code}. Response body: {error_msg}"
             ) from e
 
-        except requests.exceptions.JSONDecodeError as e:
+        except json.JSONDecodeError as e:
+            # httpx2's ``Response.json()`` raises ``json.JSONDecodeError`` when
```
acdha: I'm not sure we need the comment explaining why we're not catching a different exception — typically I'd only add that in the reverse scenario when you're explaining why a broader `except` is necessary.

acdha (dismissed):

priya-sundaram-dev:
Thanks for the review + approval, @acdha 🙏

On the JSON-decode `except`: digging into it, I think the narrow `except json.JSONDecodeError` DhavalGojiya and I landed on is actually subtly wrong here, and your instinct about the comment points right at why. This module does `import simplejson as json` when simplejson is installed, so the module-level `json` name is often **simplejson**, not the stdlib. But httpx2's `Response.json()` decodes with the **stdlib** `json` module and raises the stdlib `json.JSONDecodeError`. Those are two distinct classes (both subclass `ValueError`, but neither is a subclass of the other), so under a simplejson install `except json.JSONDecodeError` would be simplejson's class and would silently miss the stdlib error the handler is meant to wrap.

So I've reverted to the `ValueError` base class the original `requests` code used — it matches regardless of which JSON library is present — and rewritten the comment to explain *why the broader except is necessary*, which is exactly the reverse-scenario case you mentioned is worth a comment. Pushed as a follow-up. Happy to go the other way (e.g. import the stdlib `JSONDecodeError` under a distinct name) if you'd prefer to keep it narrow — just let me know which you'd rather land.

pysolr.py
DhavalGojiya: @acdha 
I am not sure why pysolr gives priority to simplejson first and then binds it as `json` in the current module namespace.

Can't we remove this simplejson check and directly use the standard `json` library instead? I think the standard `json` module has become much more powerful over time and already provides many built-in features.

Because of this `simplejson` try/except code, we can't directly use `json.JSONDecodeError` exception check here and instead need to use `ValueError`.

Or we can do something like this in pysolr.py
`from json import JSONDecodeError as StdJSONDecodeError`
And then use `StdJSONDecodeError` to catch httpx2 Json Decoder exception.

priya-sundaram-dev:
Thanks @DhavalGojiya — that's the clean fix, done in the latest push.

I went with your `from json import JSONDecodeError as StdJSONDecodeError` idea: a dedicated stdlib alias at import time, and `except StdJSONDecodeError`. Since httpx2's `Response.json()` always decodes with the stdlib `json` module, that raises the stdlib `JSONDecodeError` regardless of whether `simplejson` is installed and bound to `json` here — so the narrow catch is both specific and correct. Nicer than the broad `ValueError` I'd fallen back to.

On dropping the `simplejson` shim entirely: I agree stdlib `json` is more than capable now, but that changes public behavior (the module exports `json`, and some callers rely on simplejson's speed/`Decimal` handling), so it feels like a separate decision for the maintainers rather than something to fold into this httpx2 migration. Happy to open a follow-up PR to remove it if @acdha wants to go that way.

acdha:
> So I've reverted to the `ValueError` base class the original `requests` code used — it matches regardless of which JSON library is present — and rewritten the comment to explain _why the broader except is necessary_, which is exactly the reverse-scenario case you mentioned is worth a comment.

That's a much better reason for doing this. As long as we're raising the new exception from that, as is currently the case, the best we can do is let the caller interpret that with their knowledge of local configuration. This could even get worse if https://github.com/pydantic/httpx2/issues/965 extends support for custom serializers since that could potentially add other client's exception classes into the mix. 

This does raise the larger question of whether we still want to mess with simplejson: that dates back to when Python was just getting support for JSON in the stdlib and some pysolr users had to support those older versions. That era sailed a long time ago and there's no reason to use simplejson in the modern era. pysolr uses this in two places: the SolrCloud interface could easily be stdlib-only because it doesn't get massive documents and isn't called all that frequently, so there could be an argument for simplification there.

There are only two uses for the core pysolr functionality, we might want to reconsider whether we need to do call the JSON encoder directly at all. In the case of building documents during indexing, that appears to be due to previous versions lacking support for boosting in the JSON API, which is now gone, and removing the XML path there would make it much easier to refactor to use `httpx2.post(…, json=)`.

https://github.com/priya-sundaram-dev/pysolr/blob/b287864847831d02328d648e24d10312995e817e/pysolr.py#L850

Similarly, the use of the decoder appears to always be in places where it could easily be replaced with `response.json()`. 

The downside right now is that in the absence of [httpx2 #965](https://github.com/pydantic/httpx2/issues/965) this would be giving up the ability to use a custom decoder until that lands upstream so we might keep the current encoder/decoder implementation until that settles. It would make a great followup task to remove all of the remaining XML support since that cuts out a fair amount of code and also tends to boost performance a fair amount.

acdha (approved):

acdha:
> On dropping the `simplejson` shim entirely: I agree stdlib `json` is more than capable now, but that changes public behavior (the module exports `json`, and some callers rely on simplejson's speed/`Decimal` handling), so it feels like a separate decision for the maintainers rather than something to fold into this httpx2 migration. Happy to open a follow-up PR to remove it if @acdha wants to go that way.

I think we should. Since it'll be pysolr 4.x, we can document that in the release notes and let anyone who needs some custom feature use the existing `encoder` / `decoder` configuration options to set the library of their choice. The places where we directly call the JSON `loads` / `dumps` functions are in the ZooKeeper client where those concerns shouldn't apply.

acdha:
I merged this so we can expand testing but I think removing simplejson and evaluating the remaining XML call sites would be a great followup task.

priya-sundaram-dev:
Thanks for merging, @acdha — and agreed on the follow-up. I'll open a separate PR that drops the `simplejson` shim in favour of stdlib `json`, keeps the `encoder`/`decoder` config knobs as the supported escape hatch for anyone who needs `Decimal`/speed, and documents the change in the 4.x release notes. I'll audit the remaining XML call sites in the same pass and flag anything that deserves its own change rather than fold too much in. Will tag you when it's up.

PR #599, opened 2026-09-17
build(deps): add httpx2 alongside requests

DhavalGojiya:
## Description

We are gradually moving toward an async HTTP library, with the plan to fully replace the `requests` module with `httpx2`.

This is the initial step: it adds `httpx2` as a base dependency without touching any of the existing `requests` code.

More: https://pydantic.dev/docs/httpx2/get-started/


cclauss (approved):

acdha:
I wonder if we should plan to cut this as pysolr 4 since it's switching the underlying library in ways people might be used to customizing (e.g. injecting a Requests session with auth configuration).

DhavalGojiya:
> I wonder if we should plan to cut this as pysolr 4 since it's switching the underlying library in ways people might be used to customizing (e.g. injecting a Requests session with auth configuration).

Aren't we already in a breaking-changes state with pysolr currently, since the very next release is pysolr 4.0.0?

acdha:
> > I wonder if we should plan to cut this as pysolr 4 since it's switching the underlying library in ways people might be used to customizing (e.g. injecting a Requests session with auth configuration).
> 
> Aren't we already in a breaking-changes state with pysolr currently, since the very next release is pysolr 4.0.0?

I'm not sure if the changes to pysolr.py break existing users unless they're doing something weird calling the internal escaping function directly but I think it would be best to see this as 4.0 rather than continuing 3.x since there's definitely more room for cleanup.

DhavalGojiya:
> since there's definitely more room for cleanup.

Agree. I also have plans to add a separate, structured `CHANGELOG.md` file that documents what we changed from pysolr 3.11.0 to 4.0.0, along with at least the three previous pysolr releases and the changes introduced in each. The current `CHANGELOG.rst` file is a mess.

PR #601, opened 2026-09-18
Replace solrconfig.xml copies with config overlay in solr docker setup

DhavalGojiya:
## Description

Removed 2,627 Lines of code.

1. Remove the `solr_configs/` files from the project. We no longer need to maintain them, as we are now using Solr Config Overlay, a cool feature that allows us to override Solr configuration using just a JSON file. 
See: https://solr.apache.org/guide/solr/9_0/configuration-guide/config-api.html

2. Rename `docker-compose-solr.yml` to follow the official Docker Compose file naming convention: `compose.yaml`
When we are in the `docker/` directory and run the `docker compose` command, Docker will automatically use this file without requiring us to explicitly specify the filename.

3. Remove the network name: `search` from Network directive from the Compose file. It can conflict with developers' local networks when testing pysolr. Instead, we now use compose file -> `name: pysolr` as a project-based prefix, which should avoid such conflicts. The same applies to the container names in the Compose file.

4. Add a small README file in each Solr version directory explaining what `configoverlay.json` is and why we added things to it.

5. Optimize and simplify the `docker/solr` folder structure.

### Extra
Folder structure now
pysolr-folder


cclauss:
@priya-sundaram-dev, please review.

priya-sundaram-dev:
Nice cleanup — trading 2.6k lines of vendored `solrconfig.xml` for a small overlay is a real maintenance win, and the fact that CI is green across the whole **Solr 9 + Solr 10 × Python 3.12–3.15(t)** matrix is the strongest possible evidence the overlay actually reproduces what the old copies provided. I traced the moves end-to-end and they're consistent:

- `compose.yaml` mounts `./solr/${SOLR_VERSION:-9}.x.x/configoverlay.json`, the `solr-standalone` inline command and the relocated `docker/solr/solr-init.sh` both `cp` it into `$SITECORE_CONFIGSET/conf/configoverlay.json`, and the `test_cloud.py` docstring now points at `docker/compose.yaml`. No dangling references to `docker-compose-solr.yml`, `solr_configs/`, or `docker/scripts/` left in the tree.
- Layering only the deltas (`/mlt` on 9, `/mlt` + Tika-backed `/update/extract` on 10) over the stock `sample_techproducts_configs` is much easier to reason about than diffing two 1.2k-line XML files, and the per-version READMEs explaining *why* each handler is there (esp. the note that an overlay **replaces** a handler rather than merging, so the extract defaults are repeated) are a genuinely helpful touch.

Two small things, neither blocking:

1. **`SOLR_VERSION` does double duty** — it's both the image tag (`solr:${SOLR_VERSION:-9}`) and the overlay path (`./solr/${SOLR_VERSION:-9}.x.x/`). That's fine for bare majors (`9`, `10`), which is what CI uses, but a full value like `SOLR_VERSION=9.6.1` would resolve a valid image while the bind source `./solr/9.6.1.x.x/configoverlay.json` doesn't exist — Docker then materializes it as an empty dir and the `cp` fails confusingly. A one-line note in the docker README ("set `SOLR_VERSION` to a major only") or a small guard in `solr-docker-test-env.sh` would save a future contributor some head-scratching.
2. **Network rename wording** — worth being precise in the PR description: the `search` key is still defined and referenced by every service; what actually removes the collision risk is the new top-level `name: pysolr`, which scopes the real Docker network to `pysolr_search` instead of a bare `search`. Good fix, just not literally "removed".

Overall this reads clean and the green matrix speaks for itself. 👍

DhavalGojiya:
> That's fine for bare majors (9, 10), which is what CI uses, but a full value like SOLR_VERSION=9.6.1

Yes, that is expected. The Docker test setup is used only by our pysolr project and not by end users. We always define a major version tag because it also gives us early visibility into whether anything breaks with a new Solr release within that major version.

priya-sundaram-dev:
Makes sense — pinning a full patch version in CI to catch same-major Solr regressions early is a good reason to keep it, and since the compose setup is dev/CI-only (not shipped to end users) the `${SOLR_VERSION%.*}` path split is fine as-is. That fully resolves my only real question; the network-name wording was just a doc nit, nothing blocking. LGTM from me. Thanks for the thorough cleanup. 👍

cclauss (dismissed):

.gitignore
```diff
@@ -58,5 +58,5 @@ logs/
 
 # Apache Solr
 solr*.tgz
-solr/
-solr-app/
+/solr/
```
acdha: What problem was this trying to solve? It wasn't clear from the commit message why these needed to be changed.

priya-sundaram-dev (approved):
Reviewed and this is a really clean cut — LGTM. 👍

What I checked:
- Both `configoverlay.json` files parse as valid JSON, and the overlay is applied consistently in **both** paths: the standalone service copies it in its inline `command` (`cp $PYSOLR_CONFIG_OVERLAY …/conf/configoverlay.json` before `precreate-core`), and the SolrCloud path does the same in `solr-init.sh` before `solr zk upconfig`. So the tricky `/mlt` + `/update/extract` handlers land in every core the tests hit.
- `PYSOLR_CONFIG_OVERLAY` is correctly under `environment:` (not `volumes:`) in both services, and the read-only bind mounts line up.
- Full 14-job matrix (Solr 9/10 × 3.11–3.15 incl. free-threaded) is green, which is the real proof the overlay reproduces what the 2.6k lines of vendored `solrconfig.xml` were doing.

Deleting 2,627 lines of vendored config while keeping the upstream `solrconfig.xml` untouched is a big maintainability win, and the per-version READMEs explaining *why* the overlay has to repeat the upstream handler defaults (replace-not-merge) are exactly the kind of context future maintainers will thank you for.

One tiny, non-blocking note on the description rather than the code: item #3 says the `search` network directive was removed, but it's still declared at the bottom and referenced by each service. Functionally that's fine — the new `name: pysolr` project prefix already namespaces it to `pysolr_search`, which achieves the collision-avoidance you're after — so it's just a wording mismatch in the PR body, nothing to change in the file.

Nice work @DhavalGojiya.

.gitignore
DhavalGojiya: @acdha
We have a `docker/solr/` directory in this PR.

Due to the `.gitignore` pattern "solr/", Git is not tracking the new directory.
So, adjust this pattern to match only the project root and do not match solr/ sub directory.

acdha: Can't we remove all of these patterns now that we're only running Solr in containers? That section of the .gitignore file was covering the directories where the configuration files were stored (since relocated) and the downloaded Solr tarball.

acdha (approved):

.gitignore
DhavalGojiya: Good idea @acdha

PR #602, opened 2026-09-18
Drop optional simplejson shim in favour of stdlib json

priya-sundaram-dev:
Follow-up to #598, as discussed there with @acdha.

## What
Drop the optional `simplejson` shim and standardise on the standard-library `json` module.

Previously pysolr did:

```python
try:
    import simplejson as json  # preferred if installed
except ImportError:
    import json
```

so `pysolr.json` could resolve to either module depending on the environment. This PR makes it always the stdlib `json`.

Since `json` is now unconditionally the stdlib module, the `StdJSONDecodeError` alias I added in #598 (to catch decode failures narrowly even when `json` was bound to `simplejson`) is no longer needed — the handler catches `json.JSONDecodeError` directly.

## Compatibility
This changes public behaviour (`pysolr.json` used to be `simplejson` when installed), so it's targeted at the 4.x line and documented in the changelog under **Changes**. Anyone who needs simplejson-specific behaviour — `Decimal` handling or its C speedups — can still pass a simplejson `JSONDecoder`/`JSONEncoder` through the existing `decoder=` / `encoder=` arguments to `Solr`, which are unchanged.

There was no `simplejson` dependency declared in `pyproject.toml` (it was purely optional runtime detection), so no packaging change is needed.

## XML call sites
@acdha also flagged evaluating the remaining XML call sites as a good follow-up. I had a look: pysolr still builds update requests as XML (`_build_xml_doc` / `ElementTree`) and parses a few error/ping paths, while reads use `wt=json`. That's a larger, behaviour-affecting change (the update wire format) and deserves its own PR and discussion rather than being folded in here — happy to open one if you'd like to pursue it.

## Tests
No test changes: nothing in the suite depended on `pysolr.json` being simplejson, and the decode-error path is unchanged in behaviour (same exception type raised by `Response.json()`). `ruff check` passes.


DhavalGojiya (approved):
Brilliant.

DhavalGojiya:
https://github.com/django-haystack/pysolr/blob/f8b81febe210e6bc959e29a88e457fce3bd1b088/README.md?plain=1#L57

priya-sundaram-dev:
Good catch, thanks — done in 1451af0. Removed the `**Optional** - simplejson` line from the Requirements list since the shim is gone.

I left the `Requests 2.32.5+` line alone since `requests` is still a listed dependency in `pyproject.toml`. Happy to also refresh that section to mention httpx2 (from #598) if you'd like, but I kept it out of scope here.

acdha:
> Anyone who needs simplejson-specific behaviour — Decimal handling or its C speedups

Both of those are in the stdlib version as well — benchmarks usually have the two very close on performance since they're related code.

> pysolr still builds update requests as XML (_build_xml_doc / ElementTree) and parses a few error/ping paths

Definitely split it out separately but it shouldn't be a big deal because the Solr APIs are consistent across the serialization format. The exception was that older versions of Solr did not support boosting in the JSON API but they do now, which allows the entire _build_xml_doc method to be removed once that's added, and the various commands (commit, optimize, etc.) have JSON equivalents. 

> I left the Requests 2.32.5+ line alone since requests is still a listed dependency in pyproject.toml

I think we should remove that since it's not otherwise used and it's a dependency which will at some point have security advisories people need to patch, which is annoying if you don't use it.

priya-sundaram-dev:
Done in a516d77. Confirmed `requests` isn't imported anywhere post-#598 (`grep -rn 'import requests'` is clean — the only remaining matches are docstrings referring to HTTP requests), so I dropped it from the runtime `dependencies` in `pyproject.toml` and updated the README Requirements list to point at `httpx2` 2.13+ instead. Agreed on keeping the XML→JSON `_build_xml_doc` split as its own change — happy to pick that up separately.

CI should stay green; let me know if you'd rather I split the dependency drop into its own PR to keep the changelog clean.

DhavalGojiya:
`uv sync` required if you remove/add any dependencies in pyproject.toml file.
`uv sync` will update the uv.lock file and you need to commit that file as well otherwise our check will not pass @priya-sundaram-dev

priya-sundaram-dev:
Good catch, thanks @DhavalGojiya. Regenerated the lockfile with `uv lock` (f8fcec0) — it drops `requests` and its transitive deps (`certifi`, `charset-normalizer`, `urllib3`), leaving `idna` since httpx2 still pulls it in. Committed the updated `uv.lock` so the `UV_LOCKED=1` check passes. Also added a short changelog note for the dependency removal.

acdha (approved):

priya-sundaram-dev:
Thanks for merging, @acdha! As discussed, I'll open the `_build_xml_doc` / ElementTree → JSON update-request split as its own separate PR so it stays reviewable — I'll ping you when it's up.

PR #603, opened 2026-09-21
Enable dependency cooldowns and update dependencies

acdha:
Configure a 5 day cooldown period and confirm that this produces a clean
upgrade.

This does not tackle pre-commit hooks because we'd have to switch to
prek to implement that.


cclauss:
GitHub Actions:
https://github.com/django-haystack/pysolr/blob/896975a033296d3220715abcfaca1cfea17c0b92/.github/dependabot.yml#L14-L15

cclauss:
* #604

PR #605, opened 2026-09-21
Adopt prek for running and updating pre-commit hooks (7-day cooldown)

### LINKED ISSUE #604: Upgrade from pre-commit to prek
@priya-sundaram-dev, would you please create a pull request that adds https://prek.j178.dev for running pre-commit hooks?  We would like a 7-day cooldown period before upgrading hooks.

priya-sundaram-dev:
Closes #604.

@cclauss — this adds [prek](https://prek.j178.dev) to both **run** and **update** the pre-commit hooks, with the 7-day cooldown you asked for.

### What's here
- **`.github/workflows/prek.yml`**
  - `run` job (push / PR): runs the hooks via [`j178/prek-action`](https://github.com/j178/prek-action) — the drop-in prek replacement for the pre-commit.ci check.
  - `update` job (weekly `cron`, plus `workflow_dispatch` for an on-demand smoke test): runs `prek update --cooldown-days 7` and opens a hook-bump PR via `peter-evans/create-pull-request`. The cooldown skips hook releases newer than 7 days, so a tag that's published and yanked never lands in a PR.
- Drops the pre-commit.ci `ci:` block from `.pre-commit-config.yaml`, since prek now owns both running and updating the hooks. (pre-commit.ci can't do a cooldown; that's the gap this closes.)

### Notes
- All actions are pinned to full commit SHAs with version comments, matching `ci.yml` and keeping zizmor happy. Top-level permissions are read-only; the `update` job alone gets `contents: write` + `pull-requests: write`.
- The `update` job uses the default `GITHUB_TOKEN`. That's simplest, but PRs it opens won't themselves trigger the `run` job — fine here since the bump PR touches only `.pre-commit-config.yaml`. If you'd rather the bump PRs run the full check, I can switch it to a bot-app token.
- If you'd prefer to keep pre-commit.ci for *running* hooks and use prek only for the cooldown-aware *updates*, say the word and I'll drop the `run` job and the `ci:`-block change. I went with the full migration since a cooldown is exactly what pre-commit.ci can't offer.

Happy to adjust anything.


cclauss:
The Zizmor complaints will need to be resolved.  The first two just require adding comments that explain why those permissions are required.

acdha:
> The Zizmor complaints will need to be resolved. The first two just require adding comments that explain why those permissions are required.

That's especially interesting because zizmor zizmor 1.30.1 did not flag those when my pre-commit hooks ran locally.

acdha:
> That's especially interesting because zizmor zizmor 1.30.1 did not flag those when my pre-commit hooks ran locally.

Nevermind, this was a stale checkout on my part.

priya-sundaram-dev:
Pushed 810e61a to resolve both zizmor findings:

- **undocumented-permissions** (`contents: write` / `pull-requests: write`): added inline comments explaining each — the update job needs them so `peter-evans/create-pull-request` can push the `update-pre-commit-hooks` branch and open the PR.
- **superfluous-actions** on `create-pull-request` (informational): added a `# zizmor: ignore[superfluous-actions]` with a note, since that action wraps the branch-push + PR-open flow the runner doesn't provide natively.

@acdha re: 1.30.1 not flagging locally — CI runs the hook at zizmor v1.30.0 (pinned in `zizmor-pre-commit@v1.30.0`), and `undocumented-permissions` was promoted to a default-on audit somewhere in that range, which is likely the version skew. Verified locally with `zizmor --offline`: "No findings to report (2 suppressed)." Happy to drop the informational ignore if you'd rather leave it visible.

.github/workflows/prek.yml
```diff
@@ -0,0 +1,65 @@
...
+      - uses: j178/prek-action@4e14d07f9231acabce116ccfca13b13dd9755ece  # v3.0.0
+
+  update:
+    name: Update hooks (7-day cooldown)
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    permissions:
+      # peter-evans/create-pull-request pushes the `update-pre-commit-hooks`
+      # branch with the refreshed .pre-commit-config.yaml.
+      contents: write
```
cclauss:
```suggestion
      contents: write  # Push to the `update-pre-commit-hooks` branch
```

.github/workflows/prek.yml
```diff
@@ -0,0 +1,65 @@
...
+      # ...and opens the pull request that proposes those revisions.
+      pull-requests: write
```
cclauss:
```suggestion
      pull-requests: write  # Open a pull request to propose revisions.
```

cclauss (changes requested):

.github/workflows/prek.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      # `--cooldown-days 7` keeps hook releases newer than 7 days out of the
+      # update, to blunt supply-chain attacks that publish and yank a bad tag.
+      - name: Update hook revisions
+        run: prek update --cooldown-days 7
+
+      # create-pull-request wraps the branch push + PR open flow (with
+      # delete-branch handling) that the runner does not provide natively.
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1  # v8.1.1 # zizmor: ignore[superfluous-actions]
```
cclauss: Please do not suppress this warning.  Instead, use `gh` which is preinstalled in the runner to create the pull request.

acdha: Also, we want to minimize the number of dependencies we have, especially for things which need privileges.

priya-sundaram-dev:
Pushed 5cbc599 — dropped `peter-evans/create-pull-request` entirely. The update job now pushes the refreshed `.pre-commit-config.yaml` and opens (or refreshes) the PR with the preinstalled `gh` CLI:

- No third-party action holds `contents: write`/`pull-requests: write` anymore — only the runner's own `gh`/git.
- The `zizmor: ignore[superfluous-actions]` suppression is gone; `zizmor --offline` reports no findings, no suppressions.
- Applied your suggested wording for the two permission comments.
- `git diff --quiet` guards the no-updates case; `gh pr list --head` avoids opening a duplicate when a PR is already open (force-push refreshes the branch instead).

Thanks both for pushing on this — agreed that fewer privileged deps is the right call.

.github/workflows/prek.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - uses: j178/prek-action@4e14d07f9231acabce116ccfca13b13dd9755ece  # v3.0.0
+
+  update:
+    name: Update hooks (7-day cooldown)
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write  # Push to the `update-pre-commit-hooks` branch
+      pull-requests: write  # Open a pull request to propose revisions.
```
cclauss, line 36: Are both of these lines still required?

